PR #112, opened 2022-11-30
Importer scenario cashflow actuals

calderinisystemorph:


ifrs17-template/Files/DataNodes/DataNodes_CH.csv
```diff
@@ -3,7 +3,7 @@ ReportingNode,,,,,,
 CH,,,,,,
 ,,,,,,
 @@InsurancePortfolio,,,,,,
-SystemName,DisplayName,ContractualCurrency,LineOfBusiness,ValuationApproach,OciType,
+SystemName,DisplayName,ContractualCurrency,LineOfBusiness,ValuationApproach,OciType,Scenario,
```
amuolo, line 6: Remove Scenario from the columns of Data Node csv files because it shouldn't be mandatory. Apply to all recurrences.

ifrs17/Import/Importers.ipynb
```diff
@@ -978,35 +1011,31 @@
...
                 "\n    if(Activity.HasErrors()) return Activity.Finish();",
                 "\n    ",
                 "\n    var workspace = Workspace.CreateNew();",
-                "\n    var parsingLog = await ParseCashflowsToWorkspaceAsync(dataSet, args, workspace);",
-                "\n    if(parsingLog.Errors.Any()) return Activity.Finish().Merge(parsingLog);",
+                "\n    var log = await ParseCashflowsToWorkspaceAsync(dataSet, mainArgs, workspace);",
+                "\n    if(log.Errors.Any()) return Activity.Finish().Merge(log);",
+                "\n",
+                "\n    var targetArgs = mainArgs.Scenario == null ",
+                "\n                     ? mainArgs.RepeatOnce().Concat(await DataSource.Query<PartitionByReportingNodeAndPeriod>()  // TODO: check that concat is adding at the end of the list",
```
amuolo, line 1024: remove TODO since we have already tested this

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -18,7 +18,7 @@
         {
             "cell_type": "markdown",
             "source": [
-                "<p style=\"font-weight:bold;\"> <span style=\"font-size: 36px\"> Equality Comparers </span> </p>"
+                "<p style=\\\"font-weight:bold;\\\"> <span style=\\\"font-size: 36px\\\"> Equality Comparers </span> </p>"
```
amuolo, line 21: are these changes artifacts? if so please reset

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -91,6 +92,35 @@
...
         },
+        {
+            "cell_type": "code",
+            "source": [
+                "class YieldCurveComparer: IEqualityComparer<YieldCurve>",
+                "\n{",
+                "\n    private YieldCurveComparer(){}",
+                "\n",
+                "\n\tpublic bool Equals(YieldCurve x, YieldCurve y) => ",
+                "\n            x.Year == y.Year && x.Month == y.Month && x.Scenario == y.Scenario && x.Currency == y.Currency && x.Id == y.Id && x.Values.SequenceEqual(y.Values); // TODO: enable Precision",
```
amuolo, line 103: at the top of this notebook create a SequenceEqual static method extension for double array which accepts another double array plus a double precision. Call this method from all equality comparers in this notebook with the Precision constant.

dtrzesniak: Why do you need to also compare the Values?
Is this only used in the test, or elsewhere?

ifrs17-template/Test/ScenarioTest.ipynb
```diff
@@ -27,7 +27,27 @@
         {
             "cell_type": "code",
             "source": [
-                "#!eval-notebook \"../Initialization/InitSystemorphBaseToMemory\""
+                "#!eval-notebook \"../Initialization/InitSystemorphBaseToMemory\"",
+                "\nvar actual_EstimateTypes = new[] {\"A\",\"AA\",\"OA\",\"DA\"};",
```
dtrzesniak, line 31: Use C# standard variable names

ifrs17/Import/Importers.ipynb
```diff
@@ -377,6 +375,34 @@
             "execution_count": 0,
             "outputs": []
         },
+        {
+            "cell_type": "code",
+            "source": [
+                "public async Task<ActivityLog> CalculateAndUploadIfrsVariablesAsync(ImportArgs args, IWorkspace workspace)",
```
amuolo, line 381: Create a sub section for this import helper named "Calculate and Upload IFRS Variables"

ifrs17/Utils/Queries.ipynb
```diff
@@ -399,6 +400,74 @@
...
+                "# Raw and Ifrs Variables"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public static async Task<T[]> LoadVariablesAsync<T>(this IDataSource querySource, Guid partition) where T : BaseDataRecord",
```
dtrzesniak, line 416: Shouldn't this be something like IPartitioned?
You can make the PartitionByReportingNodeAndPeriod generic, I think and reuse this code in another context.

ifrs17/Utils/Queries.ipynb
```diff
@@ -399,6 +400,74 @@
...
+            "cell_type": "code",
+            "source": [
+                "public static async Task<T[]> LoadVariablesRelaxedAsync<T>(this IWorkspace workspace,",
+                "\n    IDataSource dataSource,",
+                "\n    Guid targetPartition,",
+                "\n    Guid defaultPartition)",
+                "\n    where T : BaseDataRecord",
+                "\n{",
+                "\n    var variables = await workspace.LoadVariablesAsync<T>(targetPartition);",
+                "\n    if (!variables.Any() && targetPartition != defaultPartition) {",
```
dtrzesniak, line 436: This check should go outside --> you always do that for all cases

dtrzesniak, line 436: This whole if cascade looks a bit fishy... you seem to be doing the 1st and 3rd statement again, but the variables might have changed in between

ifrs17/Import/Importers.ipynb
```diff
@@ -391,27 +417,23 @@
             "source": [
                 "public async Task DataNodeFactoryAsync(IDataSet dataSet, string tableName, ImportArgs args)",
                 "\n{",
-                "\n    var partition = (await DataSource.Query<PartitionByReportingNode>().Where(p => p.ReportingNode == args.ReportingNode && p.Scenario == null).ToArrayAsync()).SingleOrDefault();",
+                "\n    var partition = (await DataSource.Query<PartitionByReportingNode>().Where(p => p.ReportingNode == args.ReportingNode).ToArrayAsync()).SingleOrDefault();",
                 "\n    if(partition == null) { ApplicationMessage.Log(Error.ParsedPartitionNotFound); return; }",
```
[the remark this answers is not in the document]
amuolo: It wouldn't be possible to have more than one partition satisfying the where clause because there are listed all identity properties. Hence, I agree that we should only change SingleOrDefault -> FirstOrDefault

ifrs17-template/Files/DataNodes/DataNodes_CH.csv
calderinisystemorph: done

ifrs17/Import/Importers.ipynb
calderinisystemorph: done

ifrs17/Utils/EqualityComparers.ipynb
calderinisystemorph: We use the YieldCurveComparer to check if the newly imported Yield Curves are identical to the old ones. This setup saves computational time in case of multiple imports

ifrs17/Import/Importers.ipynb
calderinisystemorph: done

ifrs17/Import/Importers.ipynb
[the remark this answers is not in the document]
calderinisystemorph: FirstOrDefault implemented

ifrs17/Utils/EqualityComparers.ipynb
dtrzesniak: If just for comparison, use Fluent Assertions, as they have a comparer out of the box. See reportVariableTest.

ifrs17-template/Constants/CalculationEngine.ipynb
```diff
@@ -21,7 +21,8 @@
                 "var projectName = \"ifrs17\";",
                 "\nvar environmentName = \"v1.0.0\";",
                 "\nvar notebookName = \"CalculationEngine\";",
-                "\nvar calculationEngine = $\"#!import \\\"//{projectName}/{environmentName}/{notebookName}\\\"\";"
+                "\n//var calculationEngine = $\"#!import \\\"//{projectName}/{environmentName}/{notebookName}\\\"\";",
```
dtrzesniak, line 24: strange here

amuolo: this needs to be reverted

ifrs17/Utils/EqualityComparers.ipynb
amuolo: done

ifrs17/Import/Importers.ipynb
```diff
@@ -596,7 +631,10 @@
                 "\n                                                                                        FunctionalCurrency = storage.ReportingNode.Currency,",
                 "\n                                                                                        LineOfBusiness = datarow.Field<string>(nameof(DataNode.LineOfBusiness)),",
                 "\n                                                                                        ValuationApproach = datarow.Field<string>(nameof(DataNode.ValuationApproach)),",
-                "\n                                                                                        OciType = datarow.Field<string>(nameof(DataNode.OciType))",
+                "\n                                                                                        OciType = datarow.Field<string>(nameof(DataNode.OciType)),",
+                "\n                                                                                        Scenario = scenarioColumnInsurancePortfolio",
```
dtrzesniak, line 635: Is this the dimension? Scenario should not be in a dimension, only parameters and state.

ifrs17-template/Constants/CalculationEngine.ipynb
amuolo: done

amuolo:
@calderinisystemorph @dtrzesniak @dcolleoni 

All tests are green except ReportVariableTest (which is failing because of the old Scenario benchmarks). 

We use the **full union** of scenario with best estimate variables exploiting the symmetry between raw variables and ifrs variables. On top, the order of the imports doesn't matter. And this is tested in the SequenceImportTest (green)

Please review the changes done.

ifrs17/Import/Importers.ipynb
amuolo: Don't we want to support having scenario GOC?

amuolo:
@dtrzesniak @dcolleoni @calderinisystemorph 
the last issue with report benchmark has been fixed --> all tests are green

ifrs17/Import/Importers.ipynb
dtrzesniak: This should only affect the parameter (data node parameter). You can add a GIC in the dimension and control in the state if it should be active or not. The dimension should go across all scenarios, as to me this is the definition (same as for any other partition, like year, month, etc. --> if you change a Lob of a GIC, it will have an impact in all reporting periods. 

Please, make the dimension scn independent, and only allow scn for parameters.

ifrs17/Utils/Queries.ipynb
```diff
@@ -399,6 +400,66 @@
...
+                "\n    where T : IPartitioned ",
+                "\n    where P : IPartition",
+                "\n{",
+                "\n    var variables = await workspace.LoadPartitionedDataAsync<T,P>(targetPartition);",
+                "\n",
+                "\n    if(targetPartition != defaultPartition) {",
+                "\n        var isRelaxed = (format != ImportFormats.Cashflow && typeof(T).Name == nameof(IfrsVariable)) || ",
+                "\n                        (format == ImportFormats.Cashflow && typeof(T).Name == nameof(RawVariable));",
+                "\n        // For scenario re-calculation",
+                "\n        if(!variables.Any() && isRelaxed) variables = await dataSource.LoadPartitionedDataAsync<T,P>(targetPartition); ",
```
dtrzesniak, line 441: I really don't like this way of writing, as the if statement looks the same, but it's not because of the change in variables.
Can you rewrite this?

extract isRelaxed in a separated if --> if false, you don't need anything else.

await dataSource.LoadPartitionedDataAsync<T,P>(targetPartition); you need this query anyway --> extract this in a variable outside the if (var targetVariables) --> then you if this --> variables = targetVariables.Any() ?  targetVariables : await dataSource.LoadPartitionedDataAsync<T,P>(defaultPartition);

something like that

ifrs17-template/Test/Data/IfrsVariableBenchmarks/BM_CH_2020_12_MTUP10pct_A.csv
```diff
@@ -3,12 +3,630 @@ Month,ReportingNode,Scenario,Year
...
 AccidentYear,AmountType,AocType,DataNode,EconomicBasis,EstimateType,Novelty,Value
-,NIC,CF,DT2.1,,A,C,-310
-2020,NIC,CF,DT1.2,,A,C,-310
+,,CF,DT1.1,,A,C,0
 ,NIC,CF,DT1.1,,A,C,-310
+,AEA,CF,DT1.1,,A,C,-5
+,ACA,CF,DT1.1,,A,C,-10
+,PR,CF,DT1.1,,A,C,400
+,,BOP,DT1.1,,A,I,0
+,NIC,BOP,DT1.1,,A,I,0
```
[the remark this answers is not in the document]
amuolo: yes, so that we can during testing we can check that the CE produces zeros for those figures. I think it is necessary.

ifrs17-template/Test/Data/IfrsVariableBenchmarks/BM_CH_2020_12_MTUP10pct_BE.csv
```diff
@@ -3,67 +3,163 @@ Month,ReportingNode,Scenario,Year
...
-,PR,BOP,DT1.1,C,BE,N,-798.53673
-,NIC,EOP,DT1.1,L,BE,C,329.6431
-,PR,EOP,DT1.1,L,BE,C,-399.63393
-,NIC,EOP,DT1.1,C,BE,C,329.6431
-,PR,EOP,DT1.1,C,BE,C,-399.63393
-,NIC,IA,DT1.1,L,BE,N,1.01487
-,PR,IA,DT1.1,L,BE,N,-1.0972
-,NIC,IA,DT1.1,C,BE,N,1.01487
-,PR,IA,DT1.1,C,BE,N,-1.0972
+,NIC,BOP,DT1.1,L,BE,N,658.6282390585294
```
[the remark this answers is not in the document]
amuolo: I don't know, this is the output of the ExportIfrsVar notebook

ifrs17/Import/Importers.ipynb
amuolo: done

ifrs17/Utils/Queries.ipynb
amuolo: done

ifrs17-template/Test/Data/IfrsVariableBenchmarks/BM_CH_2020_12_MTUP10pct_A.csv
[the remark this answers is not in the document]
amuolo: Another option would have been to cancel zeros from the benchmark and to exclude zeros from the tests as well. However, in this way I believe we would make the test weaker because we wouldn't check whether zeros are produced or not. Hence, I think the current setup for the benchmark/tests is the safest/strongest.

ifrs17-template/Files/Dimensions.csv
```diff
@@ -119,7 +119,7 @@ LR,Loss Recovery Component,40,AoC,7,NotApplicable,,,,,,
 PL,Profit and Loss,50,None,7,NotApplicable,,,,,,
 AA,Advance Actuals,60,AoC,3,NotApplicable,PayablePR,ReceivableNIC,ReceivableICO,RiReceivablePR,RiPayableNIC,
 OA,Overdue Actuals,70,AoC,3,NotApplicable,ReceivablePR,PayableNIC,PayableICO,RiPayablePR,RiReceivableNIC,
-DA,Deferrable Actuals,80,AoC,3,NotApplicable,,,,,,
+DA,Deferrable Actuals,80,AoC,7,NotApplicable,,,,,,
```
dcolleoni, line 122: @tkukuljan check this out for your test set up

ifrs17-template/Test/Data/ExportReportVariables.ipynb
```diff
@@ -174,7 +174,7 @@
         {
             "cell_type": "code",
             "source": [
-                "var scopesData = (await GetScopesDataAsync(args, reportStorage, Scopes)).Select(rv => rv with { Value = Math.Round(rv.Value, roundingDigits)});",
+                "var scopesData = (await GetScopesDataAsync(args, reportStorage, Scopes)).AggregateBy().Select(rv => rv with { Value = Math.Round(rv.Value, roundingDigits)});",
```
dcolleoni, line 177: why is this needed?

ifrs17-template/Test/Data/IfrsVariableBenchmarks/BM_CH_2020_12_MTUP10pct_BE.csv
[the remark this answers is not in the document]
dcolleoni: it shouldn't be at some point we had a code to take on 5 digits. 
We should put it back in place and generate BM accordingly. 
Now here it is not clear what is changing - only number of digits or the number? -

dcolleoni:
There are BM full of 0s. Are those all variables stored in the DB? It is really too much, some optimization should be in place.

ifrs17-template/Test/IfrsVariablesTest.ipynb
```diff
@@ -184,40 +188,22 @@
...
-                "\n                                                    x.EconomicBasis == variable.EconomicBasis && ",
-                "\n                                                    x.AccidentYear == variable.AccidentYear",
-                "\n                                                     )?.Value ?? 0.0d;",
-                "\n          errors.Add(new BenchmarkTestResult(\"Value does not match for Variable: \" + message, -bmValue, computedValue));",
-                "\n        }",
+                "\n        foreach (var variable in computedNotExpected)",
+                "\n            errors.Add(new BenchmarkTestResult(\"Extra computed variable for: Partition \" + variable.Partition + \", \" + variable.ToIdentityString()));",
+                "\n",
+                "\n    var comparer = EqualityComparer<IfrsVariable>.Instance; ",
+                "\n    foreach (var a in expected) {",
```
dcolleoni, line 201: improve the name of the variables a and b make the code hard to read

ifrs17-template/Test/ScenarioDataImportTest.ipynb
```diff
@@ -0,0 +1,868 @@
...
+                "DataSource.Reset(x => x.ResetType<RawVariable>().ResetType<IfrsVariable>());"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "var actualEstimateTypes = new[] {\"A\",\"AA\",\"OA\"};"
```
dcolleoni, line 48: Use GetImportActualEstimateType instead

ifrs17-template/Test/ScenarioDataImportTest.ipynb
```diff
@@ -0,0 +1,868 @@
...
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "var cashflowsScenario = @\"",
+                "\n@@Main",
+                "\nReportingNode,Year,Month,Scenario",
+                "\nCH,2020,12,Test",
```
dcolleoni, line 157: Do we have a validation in the importer that checks that the value of Scenario must be within the list of the imported Scenarios?

ifrs17-template/Test/ScenarioDataImportTest.ipynb
```diff
@@ -0,0 +1,868 @@
...
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "diffs.Where(x => actualEstimateTypes.Contains(x.EstimateType)).ToArray().Length.Should().NotBe(0);",
+                "\ndiffs.Where(x => x.EstimateType == EstimateTypes.CU).ToArray().Length.Should().Be(0);",
```
dcolleoni, line 248: why do you check only for CU? I would check for all EstimateType that are not related to Actuals - BE and Ra for example.

ifrs17-template/Test/ScenarioDataImportTest.ipynb
```diff
@@ -0,0 +1,868 @@
...
+                "\nvar diffs = ivsScenarioCashflowUpdateCU.Except(ivsScenarioInitActuals, IfrsVariableComparer.Instance());"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "diffs.Where(x => x.EstimateType == EstimateTypes.DA).ToArray().Length.Should().NotBe(0);",
```
dcolleoni, line 311: I would add : 
diffs.Where(x => x.EstimateType == EstimateTypes.C || x.EstimateType == EstimateTypes.L).ToArray().Length.Should().NotBe(0);

ifrs17-template/Test/ScenarioDataImportTest.ipynb
```diff
@@ -0,0 +1,868 @@
...
+                "\nivsScenarioActualNoACAAEA.Where(x => x.EstimateType == EstimateTypes.F && Math.Abs(x.Value) > Precision).Count().Should().NotBe(0);",
```
dcolleoni, line 431: I woudl remove Math.Abs(x.Value) > Precision 
there should not be any change for F not even variables with 0 value

ifrs17-template/Test/ScenarioDataImportTest.ipynb
```diff
@@ -0,0 +1,868 @@
...
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "await DataSource.Partition.SetAsync<PartitionByReportingNodeAndPeriod>(argsScenario);",
+                "\nvar ivsScenarioReimportEqualToBestEstimate = await DataSource.Query<IfrsVariable>().ToArrayAsync(); ",
+                "\nvar diffs = ivsScenarioReimportEqualToBestEstimate.Except(ivsBestEstimateInit, IfrsVariableComparer.Instance()).Where(x => Math.Abs(x.Value) > Precision);"
```
dcolleoni, line 505: Also here, if we save the 0s correctly we should not have any difference: 
remove .Where(x => Math.Abs(x.Value) > Precision)

ifrs17-template/Test/ScenarioDataImportTest.ipynb
```diff
@@ -0,0 +1,868 @@
...
+                "# Reimport of Best Estimate Cashflow: zero impact on Scenario"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "markdown",
+            "source": [
+                "Reimport of best estimate with different gross cash flow, whereas reinsurance business is the same of the initialization. This causes an update of the best estimate IFRS variables but it must have no impact on Scenario variables because gross cash flow for Scenario have already been imported. "
```
dcolleoni, line 532: improve the description as it is not clear that the BE var that will be imported have the same identity property of those imported for the scemario.

ifrs17-template/Test/Data/IfrsVariableBenchmarks/BM_CH_2020_12_MTUP10pct_A.csv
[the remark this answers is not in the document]
dtrzesniak: It is just that we are using computation time to compare millions of 0. Yes, it is safe, I agree, but maybe there is a better solution to achieve the same, since this will be slow and will make BM files difficult to read.

ifrs17/Import/ImportScopeCalculation.ipynb
```diff
@@ -2053,26 +2053,27 @@
...
-                "\n                                                                                                                                          AccidentYear = x.Identity.AccidentYear,",
-                "\n                                                                                                                                          AmountType = x.Identity.AmountType,",
-                "\n                                                                                                                                          Value = x.Value,",
-                "\n                                                                                                                                          Partition = GetStorage().TargetPartition",
-                "\n                                                                                                                                        });",
+                "\n    IEnumerable<IfrsVariable> PvLocked => GetScope<PvLocked>(Identity).PresentValues.Select(x => ",
```
dcolleoni, line 2056: I prefer previous indentation. It also ease the review. Could you go back? thanks

ifrs17-template/Test/Data/IfrsVariableBenchmarks/BM_CH_2020_12_MTUP10pct_A.csv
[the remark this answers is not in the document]
calderinisystemorph: I hve createad an issue for this problem(https://github.com/Systemorph/IFRS17CalculationEngine/issues/132). We are aware that the current implementation is suboptimal, but the topic is quite delicate and need a dedicated PR,

ifrs17/Import/ImportStorage.ipynb
```diff
@@ -273,15 +274,8 @@
                 "\n",
                 "\n        if(DefaultPartition != TargetPartition) {",
                 "\n            await querySource.Partition.SetAsync<PartitionByReportingNodeAndPeriod>(DefaultPartition);",
-                "\n            var defaultRawVariables = await querySource.Query<RawVariable>()",
-                "\n                                                       .Where(rv => rv.Partition == DefaultPartition)",
```
dcolleoni, line 277: why is this Where removed? Isn't it necessary for physical DB?

ifrs17/Import/ImportStorage.ipynb
```diff
@@ -273,15 +274,8 @@
...
-                "\n                                                       .Where(rv => primaryScope.Contains(rv.DataNode)).ToArrayAsync();",
-                "\n            var defaultIfrsVariables = await querySource.Query<IfrsVariable>()",
-                "\n                                                        .Where(iv => iv.Partition == DefaultPartition)",
```
dcolleoni, line 280: same here

ifrs17-template/Test/ScenarioDataImportTest.ipynb
calderinisystemorph: proposed improvement: Reimport of best estimate with different gross cash flow, whereas reinsurance business is the same of the initialization. This causes an update of the best estimate IFRS variables but it must have no impact on Scenario variables because gross cash flow contains only a combination of AocType, Novelty and AmoutType that have already been imported for Scenario (see above the variable cashflowsScenarioEqualToBestEstimate).

ifrs17-template/Test/ScenarioDataImportTest.ipynb
calderinisystemorph: done

ifrs17-template/Test/ScenarioDataImportTest.ipynb
calderinisystemorph: here we are comparing Scenario IfrsVariables with BestEstimate ones. If we don't exclude zeros, the test fails since it cannot match any zeros.

ifrs17-template/Test/Data/ExportReportVariables.ipynb
amuolo: because the import scopes are generating actuals for all novelties (N+I+C), while the report scopes are setting each one of the queried ReportVariable to Novelty C. This means at report level there are duplicates, that is, ReportVariable with the same id props but different values. Of course, those coming from actuals with Novelty N and I are zeros (default). Finally, this AggregateBy() allows to discard the duplicates (zeros) making the scopes data equal to the benchmark data that is imported from file and thus has no duplicate (because of how import works).

ifrs17/Import/ImportStorage.ipynb
amuolo: yes, you are right, this needs to be there to support physical db

ifrs17/Import/ImportStorage.ipynb
amuolo: yes

ifrs17/Import/ImportScopeCalculation.ipynb
amuolo: Properties weren't indented at all, and with the removal of the Where filter the format of these cells becomes even worse. A refactor of the indentation was necessary.

ifrs17-template/Test/ScenarioDataImportTest.ipynb
calderinisystemorph: The test section about CU was written to prove that the changes in Dimensions.csv from "DA,Deferrable Actuals,80,AoC,3,NotApplicable,,,,,," to "DA,Deferrable Actuals,80,AoC,7,NotApplicable,,,,,," was working properly. More specifically, we expected a chian of changes from CU => AM => DA. We also add C, L And F for completeness.  Testing all EstimateType not related to Actuals seems to me a bit too generic test. I would rather create a new issue to improve the test coverage in another notebook.

ifrs17/Import/ImportStorage.ipynb
```diff
@@ -273,15 +274,8 @@
...
-                "\n                                                        .Where(iv => primaryScopeFromParsedVariables.Contains(iv.DataNode)",
-                "\n                                                                     && !EstimateTypesByImportFormat[ImportFormat].Contains(iv.EstimateType)  ",
-                "\n                                                                     || primaryScopeFromLinkedReinsurance.Contains(iv.DataNode) ",
-                "\n                                                                     || secondaryScope.Contains(iv.DataNode)).ToArrayAsync();",
+                "\n            var defaultRawVariables = await querySource.Query<RawVariable>().Where(rv => rv.Partition == DefaultPartition).Where(rv => primaryScope.Contains(rv.DataNode)).ToArrayAsync();",
+                "\n            var defaultIfrsVariables = await querySource.Query<IfrsVariable>().Where(iv => iv.Partition == DefaultPartition).Where(iv => allImportScopes.Contains(iv.DataNode)).ToArrayAsync();         ",
```
calderinisystemorph, line 278: @dcolleoni  Instead of saving all zeros, we need to store only 0 value variable that are in the intersection between the BE and the computed SENS ifrsVariable sets

amuolo: why do you write this comment here in the ImportStorage?

ifrs17/Utils/Queries.ipynb
```diff
@@ -399,6 +399,67 @@
...
+            "source": [
+                "public static async Task<T[]> RelaxedQueryPartitionedDataAsync<T,P>(this IWorkspace workspace, IDataSource dataSource, Guid targetPartition, Guid defaultPartition, string format)",
+                "\n    where T : IPartitioned ",
+                "\n    where P : IPartition",
+                "\n{",
+                "\n    var isRelaxed = (format != ImportFormats.Cashflow && typeof(T).Name == nameof(IfrsVariable)) || ",
+                "\n                    (format == ImportFormats.Cashflow && typeof(T).Name == nameof(RawVariable));",
+                "\n",
+                "\n    var variablesFromWorkspace = await workspace.LoadPartitionedDataAsync<T,P>(targetPartition);",
+                "\n    if(targetPartition == defaultPartition || !isRelaxed || variablesFromWorkspace.Any()) return variablesFromWorkspace; ",
```
dcolleoni, line 438: I find this bool statement confusing: 
in the case the bools in this if are false,false,true => is sens, is relaxed, there are variable in the workspace, then we return the variables from the workspace. In fact the result is not relaxed. So why was the isRelaxed variable true?

ifrs17/Utils/Queries.ipynb
```diff
@@ -399,6 +399,67 @@
...
+                "\n",
+                "\n    // For scenario re-calculation",
+                "\n    var variablesFromDataSource = await dataSource.LoadPartitionedDataAsync<T,P>(targetPartition);",
+                "\n    if(variablesFromDataSource.Any()) return variablesFromDataSource;",
+                "\n",
+                "\n    // For scenarios related to parameters solely",
+                "\n    var variablesDefault = await dataSource.LoadPartitionedDataAsync<T,P>(defaultPartition);",
+                "\n    await workspace.Partition.SetAsync<P>(targetPartition);",
```
dcolleoni, line 446: I would reset the partition in the LoadPartitionedDataAsync method. 
In there you can: 
1. GetCurrentPartition
2. set the partition 
3. reset to the initial partition

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -18,7 +18,7 @@
         {
             "cell_type": "markdown",
             "source": [
-                "<p style=\"font-weight:bold;\"> <span style=\"font-size: 36px\"> Equality Comparers </span> </p>"
+                "<p style=\"font-weight:bold;\"> <span style=\"font-size: 36px\"> Equality Comparers </span> </p"
```
dcolleoni, line 21: this was most probably an unwanted change. Pls reset.

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -27,7 +27,33 @@
...
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public static bool SequenceEqual(this double[] defaultArray, double[] testArray, double precision)",
+                "\n{",
+                "\n    if (defaultArray.Length != testArray.Length) return false;",
```
dcolleoni, line 51: add a check for null

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -46,16 +72,11 @@
                 "\n",
                 "\n    public bool Equals(RawVariable x, RawVariable y) =>",
                 "\n        x.AccidentYear == y.AccidentYear && x.AmountType == y.AmountType && x.DataNode == y.DataNode && x.AocType == y.AocType && ",
-                "\n        x.Novelty == y.Novelty && x.EstimateType == y.EstimateType && (IgnoreValues ? true : x.Values.SequenceEqual(y.Values)); // TODO: enable Precision",
+                "\n        x.Novelty == y.Novelty && x.EstimateType == y.EstimateType && (IgnoreValues ? true : x.Values.SequenceEqual(y.Values, Precision));",
                 "\n",
-                "\n    public int GetHashCode(RawVariable v) => IgnoreValues  ",
-                "\n                                            ? HashCode.Combine(v.AccidentYear, v.AmountType, v.DataNode, v.AocType, v.Novelty) ",
-                "\n                                            : HashCode.Combine(v.AccidentYear, v.AmountType, v.DataNode, v.AocType, v.Novelty, v.Values);",
+                "\n    public int GetHashCode(RawVariable v) => 0;",
```
dcolleoni, line 77: Is this correct?

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -75,22 +96,37 @@
...
+            "cell_type": "code",
+            "source": [
+                "class YieldCurveComparer: IEqualityComparer<YieldCurve>",
+                "\n{",
+                "\n    private YieldCurveComparer(){}",
+                "\n",
+                "\n\tpublic bool Equals(YieldCurve x, YieldCurve y) => ",
+                "\n        x.Year == y.Year && x.Month == y.Month && x.Scenario == y.Scenario && x.Currency == y.Currency && x.Id == y.Id && x.Values.SequenceEqual(y.Values, Precision);",
+                "\n\t",
+                "\n\tpublic int GetHashCode (YieldCurve x) => 0;",
```
dcolleoni, line 120: also here

dcolleoni:
We should introduce Toggle switch to unable disable certain functionalities in the code like Sens. We can do it in the next pr but before next deployment.

ifrs17/Utils/Queries.ipynb
amuolo: done

ifrs17/Utils/Queries.ipynb
amuolo: this is now a lot improved! Please check

ifrs17/Utils/EqualityComparers.ipynb
amuolo: yes, being the equality comparer based on a identity property matching function, the hash code is not used here. Any proper method to compute hash would only sacrifice performance without added value.

ifrs17/Utils/EqualityComparers.ipynb
amuolo: see comment above

ifrs17-template/Test/Data/IfrsVariableBenchmarks/BM_CH_2020_12_MTUP10pct_BE.csv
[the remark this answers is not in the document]
amuolo: done

ifrs17/Utils/EqualityComparers.ipynb
calderinisystemorph: Check added, let me know if that's ok

ifrs17-template/Test/ScenarioDataImportTest.ipynb
calderinisystemorph: The Validation Topic has been discussed with @dcolleoni: the implementation depends on the exposure of the complete Validator at Platform level, and therefore postponed.

ifrs17/Import/ImportStorage.ipynb
amuolo: this is now implemented

ifrs17-template/Test/ScenarioDataImportTest.ipynb
dcolleoni: proceed as discussed.

dcolleoni (approved):

ifrs17/Import/Importers.ipynb
```diff
@@ -586,88 +632,83 @@
                 "\n    await storage.InitializeAsync();",
                 "\n    if(Activity.HasErrors()) return Activity.Finish();",
                 "\n       ",
-                "\n    var errors = new List<string>();",
+                "\n    var errors = new List<string>();   ",
```
dtrzesniak, line 635: Is this used?

ifrs17/Import/Importers.ipynb
```diff
@@ -951,19 +993,22 @@
...
                 "\n                ",
                 "\n            var item = new RawVariable {",
                 "\n                DataNode = dataNode,",
                 "\n                AocType = aocType,",
                 "\n                Novelty = novelty,",
                 "\n                AmountType = amountType,",
                 "\n                EstimateType = estimateType,",
-                "\n                AccidentYear = dataset.Tables[ImportFormats.Cashflow].Columns.Where(x => x.ColumnName == nameof(RawVariable.AccidentYear)).Count()>0?",
-                "\n                            (Int32.TryParse((datarow.Field<string>(nameof(RawVariable.AccidentYear))), out var accidentYear)? accidentYear : (int?)null) : (int?)null, ",
+                "\n                AccidentYear = accidentYearColumn? (Int32.TryParse((datarow.Field<string>(nameof(RawVariable.AccidentYear))), out var accidentYear)? accidentYear : (int?)null) : (int?)null, ",
```
dtrzesniak, line 1011: Can we improve this line? This double (int?) is not nice.

ifrs17/Import/Importers.ipynb
```diff
@@ -984,35 +1029,29 @@
...
+                "\n    var mainArgs = (await GetArgsFromMainAsync<PartitionByReportingNodeAndPeriod>(dataSet)) with {ImportFormat = ImportFormats.Cashflow};",
                 "\n    if(Activity.HasErrors()) return Activity.Finish();",
+                "\n",
+                "\n    var targetArgs = mainArgs.Scenario == null ",
+                "\n                     ? mainArgs.RepeatOnce().Concat(await DataSource.Query<PartitionByReportingNodeAndPeriod>() ",
+                "\n                               .Where(x => x.ReportingNode == mainArgs.ReportingNode && x.Year == mainArgs.Year && x.Month == mainArgs.Month && x.Scenario != null)",
+                "\n                               .Select(x => new ImportArgs(x.ReportingNode, x.Year, x.Month, default(Periodicity), x.Scenario, ImportFormats.Cashflow)).ToArrayAsync())",
+                "\n                     : mainArgs.RepeatOnce();",
+                "\n",
+                "\n    if (targetArgs.Count() > 1) ApplicationMessage.Log(Warning.ScenarioReCalculations, String.Join(\", \", targetArgs.Skip(1).Select(x => x.Scenario)));",
```
dtrzesniak, line 1041: How can you guarantee the order? Skip 1 is not what you want, but something like First(x => x.Scenario is not null).

Also you have the if already on line 1035, why do you need to count targetArgs? Issue the Warning there.

ifrs17/Import/Importers.ipynb
```diff
@@ -1106,31 +1145,29 @@
...
+                "\n                               .Where(x => x.ReportingNode == mainArgs.ReportingNode && x.Year == mainArgs.Year && x.Month == mainArgs.Month && x.Scenario != null)",
+                "\n                               .Select(x => new ImportArgs(x.ReportingNode, x.Year, x.Month, default(Periodicity), x.Scenario, ImportFormats.Cashflow)).ToArrayAsync())",
+                "\n                     : mainArgs.RepeatOnce();",
                 "\n",
-                "\n    var universe = Scopes.ForStorage(storage).ToScope<IModel>();",
-                "\n    var identities = universe.GetScopes<GetIdentities>(storage.DataNodesByImportScope[ImportScope.Primary]).SelectMany(s => s.Identities);",
-                "\n    var ivs = universe.GetScopes<ComputeIfrsVarsActuals>(identities).SelectMany(x => x.CalculatedIfrsVariables);",
-                "\n    if(Activity.HasErrors()) return Activity.Finish().Merge(parsingLog);",
+                "\n    if (targetArgs.Count() > 1) ApplicationMessage.Log(Warning.ScenarioReCalculations, String.Join(\", \", targetArgs.Skip(1).Select(x => x.Scenario)));",
```
dtrzesniak, line 1157: Same as above

dtrzesniak: So you need this code repetition?

dtrzesniak (approved):

ifrs17-template/Test/Data/ExportIfrsVariable.ipynb
```diff
@@ -107,22 +107,13 @@
                 "\n                                                                    .WithColumn(x => x.ImportFormat, x => x.Delete())",
                 "\n                                                    )",
                 "\n                                    .WithTable<IfrsVariable>(tableConfig => tableConfig",
-                "\n                                                                  .WithSource(source => computedVariables[estimateType].AsQueryable())",
+                "\n                .WithSource(source => computedVariables[estimateType].Select(x => x with {Value = Math.Round(x.Value, 5)}).AsQueryable())",
```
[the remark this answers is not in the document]
calderinisystemorph: implemented

ifrs17/Import/Importers.ipynb
calderinisystemorph: The order is guaranteed by the construction of the query in line 1036 (using of Concat()).  In line 1041 I need to collect all Scenario to produce a string to Warning Message.   
First(x => x.Scenario is not null) would return only one Scenario

ifrs17/Import/Importers.ipynb
calderinisystemorph: I can implement a refactoring in the next PR since I need to implement it also in YieldCurve Importer PR

ifrs17-template/Test/Data/ExportIfrsVariable.ipynb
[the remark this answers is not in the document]
dtrzesniak: The 5 is still hardcoded. You must use the BMPrecison variable

ifrs17/Import/Importers.ipynb
dtrzesniak: You can also use a Where() if you need more.

If mainArgs.Scenario == null, you have mainArgs.RepeatOnce(), which will have count == 1, as you don't add anything else.

All other cases seem to have count > 1, unless the Where in 1037 is empty (can it be?). Extract this scenario query in in a variable and this is the only thing you need to build the Warning --> you don't need to Concat and then separate/Skip in the next line.

You can do all this logic inside and if where mainArgs.Scenario == null

on line 1032, you can already have an array (this RepeatOnce).

ifrs17/Import/Importers.ipynb
calderinisystemorph: Proposal: 
change     
var accidentYearColumn = dataSet.Tables[ImportFormats.Cashflow].Columns.Where(x => x.ColumnName == nameof(RawVariable.AccidentYear)).Count() > 0;

=>
var missingAccidentYearColumn = dataSet.Tables[ImportFormats.Cashflow].Columns.Where(x => x.ColumnName == nameof(RawVariable.AccidentYear)).Count() == 0; 

and now: 
AccidentYear = missingAccidentYearColumn
                                ? (int?)null 
                                : Int32.TryParse((datarow.Field<string>(nameof(RawVariable.AccidentYear))), out var accidentYear)
                                    ? accidentYear 
                                    : (int?)null,

ifrs17/Import/Importers.ipynb
calderinisystemorph: I agree, it is not used anymore.  Deleted

ifrs17/DataModel/DataStructure.ipynb
```diff
@@ -2070,7 +2070,43 @@
         {
             "cell_type": "code",
             "source": [
-                "public record Args(string ReportingNode, int Year, int Month, Periodicity Periodicity, string Scenario){}"
+                "public record Args",
```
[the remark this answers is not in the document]
calderinisystemorph: added to my personal TODO list

ifrs17/Import/Importers.ipynb
dtrzesniak: What about this for the last part:
AccidentYear = (!missingAccidentYearColumn && Int32.TryParse((datarow.Field(nameof(RawVariable.AccidentYear))), out var accidentYear))
? accidentYear
: default

==> default should work, I think, because AccidentYear is defined to be int?. Otherwise, use (int?)null

dcolleoni: 2 ternary operators are probably not needed.
can you say the following?
hasAccidentYearColumn && Int32.TryParse((datarow.Field(nameof(RawVariable.AccidentYear))), out var accidentYear) 
?  accidentYear
: something;

calderinisystemorph: implemented

ifrs17/Utils/EqualityComparers.ipynb
calderinisystemorph: Issue created: https://github.com/Systemorph/IFRS17CalculationEngine/issues/146

ifrs17/Import/Importers.ipynb
calderinisystemorph: implemented

PR #119, opened 2022-12-05
Split compute all scopes

nnikolopoulos:


ifrs17/Import/ImportScopeCalculation.ipynb
```diff
@@ -2174,7 +2203,31 @@
...
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public interface TmToIfrsVariable: IScope<ImportIdentity, ImportStorage>",
+                "\n{",
+                "\n    private string EconomicBasis => Identity.ValuationApproach == ValuationApproaches.VFA ? EconomicBases.C : EconomicBases.L;",
+                "\n    protected IEnumerable<IfrsVariable> amortizationFactor => Identity.AocType == AocTypes.AM",
```
nnikolopoulos, line 2218: A point worth noticing is that I changed the access modifier here from **private** to **protected**

sbusato: This is fine, otherwise you cannot access the amortizationFactor in the scopes below.

dcolleoni (approved):

PR #143, opened 2022-12-20
Moving tests to engine

tkukuljan:
Adapting tests (AocStructureTest, ImportStorageTest, ReportStorageTest, TechnicalMarginTest) so that they can be moved to the engine.


ifrs17-template/Test/TestData.ipynb
```diff
@@ -0,0 +1,366 @@
...
+            "name": "C#"
+        }
+    },
+    "nbformat": 4,
+    "nbformat_minor": 5,
+    "cells": [
+        {
+            "cell_type": "code",
+            "source": [
+                "#!import \"../../ifrs17/CalculationEngine\""
```
dcolleoni, line 21: You can't import CE in this way.

ifrs17-template/Test/TestData.ipynb
```diff
@@ -0,0 +1,366 @@
...
+                "# Variables"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "var groupOfInsuranceContracts = \"DT1.1\"; //TODO",
```
dcolleoni, line 202: what's the TODO? remove if any

ifrs17-template/Test/TestData.ipynb
```diff
@@ -0,0 +1,366 @@
...
+                "\n    ContractualCurrency =  \"USD\",",
+                "\n    LineOfBusiness = \"ANN\",",
+                "\n    ValuationApproach = \"BBA\",",
+                "\n    OciType =\"Default\" ,",
+                "\n    SystemName = \"DT1.1\",",
+                "\n    DisplayName = \"DT1.1 OCI LRC PA 0.8\"",
+                "\n};",
+                "\n",
+                "\nvar dtr11 = new GroupOfReinsuranceContract(){",
+                "\n    Portfolio = \"DT1\",",
```
dcolleoni, line 286: Group of Reinsurance Contracts belong to Reinsurance Portfolio.
Here should be DTR1 which needs to be created

ifrs17-template/Test/TestData.ipynb
```diff
@@ -0,0 +1,366 @@
...
+                "# Projection Configuration"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "var projcetionConfigurationPrevious = new ProjectionConfiguration(){",
```
dcolleoni, line 134: why do you need projection configuration?
Are you sure these values correspond to P11? 
Do you actually need P11?

tkukuljan: In the TechnicalMarginTest, some scopes (ultimately TechnicalMarginForIA) call ImportStorage.GetTimeStep(0).
I double checked that values are ok (P0 is end of january).
GetTimeStep(0) calls ProjectionConfiguration[0], and ProjectionConfiguration is populated with the following query in the  ImportStorage.InitializeAsync():
(await workspace.Query<ProjectionConfiguration>()
                                               .Where(x => x.Shift > 0 || x.TimeStep == args.Month || (x.TimeStep > args.Month && x.TimeStep % periodicityInMonths == 0))
                                               .OrderBy(x => x.Shift).ThenBy(x => x.TimeStep).ToArrayAsync());
From all the configurations (0,12) seemed to be the one I wanted, as args.Month is 12 in the test case.

ifrs17-template/Test/TestData.ipynb
tkukuljan: Later on when the file will be moved in the CE, it can be done in this way.
Do you think it is fine to keep it like this, or change to the one from the Constants folder?

ifrs17-template/Test/TestData.ipynb
tkukuljan: Removed

ifrs17-template/Test/TestData.ipynb
tkukuljan: Created

ifrs17-template/Test/TestData.ipynb
dcolleoni: This code does not work in the real set up. It works only on your local project where you upload everything in one project. I don't know exactly what you need here... but if it is not clear how to proceed, you should work with 3 projects instead.

dcolleoni: This code does not work in the real set up. It works only on your local project where you upload everything in one project. I don't know exactly what you need here... but if it is not clear how to proceed, you should work with 3 projects instead.

ifrs17-template/Test/TestData.ipynb
dcolleoni: In that test you don't hit that part of the code. So I guess any value would work. I suggest anyway to put all entries, not only one.

ifrs17-template/Test/TestData.ipynb
tkukuljan: I want the "CalculationEngine" notebook to be imported from the CE. 
How would you import it here?

ifrs17-template/Test/TestData.ipynb
tkukuljan: Did in the recent commit

dcolleoni (approved):

PR #151, opened 2022-12-21
Adding mandatory AoC step exception in the filter 

tkukuljan:
and adding warnings for not importing mandatory AoC steps.


ifrs17/Import/Importers.ipynb
```diff
@@ -509,6 +487,99 @@
...
+            "cell_type": "code",
+            "source": [
+                "public async static Task ValidateForMandatoryAocSteps(this IWorkspace workspace, IDataSet dataSet, HashSet<AocStep> mandatoryAocSteps)",
+                "\n{",
+                "\n    var missingAocStepsByIdentityProperties = (await workspace.Query<RawVariable>().ToListAsync())",
+                "\n                     .GroupBy(x => IdentityPropertyReader<RawVariable>.Instance.GetIdentityString(x),",
+                "\n                              x => new AocStep(x.AocType, x.Novelty),",
+                "\n                              (properties, aocSteps) => (Properties: properties, AocSteps: mandatoryAocSteps.Except(aocSteps))",
+                "\n                             );",
+                "\n    foreach(var aocSteps in missingAocStepsByIdentityProperties) ",
```
amuolo, line 575: aocSteps --> missingSteps

ifrs17/Import/Importers.ipynb
```diff
@@ -509,6 +487,99 @@
...
+                "\n        foreach(var aocStep in aocSteps.AocSteps)  ApplicationMessage.Log(Warning.MandatoryAocStepMissing, aocStep.AocType, aocStep.Novelty, aocSteps.Properties);",
```
amuolo, line 576: foreach(var aocStep in aocSteps.AocSteps) --> foreach(var missingStep in missingSteps)
+
change the corresponding variables in the ApplicationMessage

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -183,6 +183,52 @@
             "metadata": {},
             "execution_count": 0,
             "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public class IdentityPropertyReader<T> where T:BaseVariableIdentity {",
```
amuolo, line 190: Please move this class in a separate notebook - I would say a new one named IdentityPropertyReader - since it doesn't fit this notebook EqualityComparers

ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -183,6 +183,52 @@
...
+                "\n    private static Func<T, string> identityReader;",
+                "\n    ",
+                "\n    private IdentityPropertyReader(){",
+                "\n        identityReader = GetIdentityReader();",
+                "\n    }",
+                "\n    public static readonly IdentityPropertyReader<T> Instance = new IdentityPropertyReader<T>();",
+                "\n",
+                "\n    private Func<T, string> GetIdentityReader(){",
+                "\n        var parameter = Expression.Parameter(typeof(RawVariable));",
```
amuolo, line 199: typeof(RawVariable) --> typeof(T)
apply this change to all other occurrences

ifrs17/Import/Importers.ipynb
tkukuljan: this one i changed to parsedAocSteps. I would keep it if you confirm

ifrs17/Utils/IdentityPropertyReader.ipynb
```diff
@@ -0,0 +1,80 @@
...
+                "\nusing static Systemorph.Vertex.Equality.IdentityPropertyExtensions;"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public class IdentityPropertyReader<T> where T:BaseVariableIdentity {",
```
amuolo, line 32: We need to improve this method: 
- allowing the user to be able to set the excluded properties and get the corresponding eventually updated instance;
- simplifying the concat logic.
We can work together to achieve both these goals.

ifrs17/Utils/IdentityPropertyReader.ipynb
```diff
@@ -0,0 +1,69 @@
...
+                "\n        var key = string.Join(\",\", excludingProperties.OrderBy(x=>x));",
+                "\n        if(!InstancesByExcludedProperties.TryGetValue(key, out var instance))",
+                "\n            return InstancesByExcludedProperties[key] = new IdentityPropertyReader<T>(excludingProperties);",
+                "\n        return instance;",
+                "\n    }",
+                "\n    public string GetIdentityString(T x) => identityReader(x);",
+                "\n    ",
+                "\n    public static string Concat(string first, string second) => first + \" \" + second;",
+                "\n    public static string GetString(Object item) => (item == null) ? \"\" : item.ToString();",
+                "\n}"
```
dtrzesniak, line 62: I don't like this method:
 1. this is very generic, and should not belong in a client project
 2. this has been implemented in other places before: platform V7, IFRS Generic V7 and it would be yet another version for the same functionality
 3. I think a few concepts must be discuss with the dev team, as they have already implemented similar concepts (see 2)
 4. I think this is not thread safe, with your static dictionary (line 41)
 5. Why do you reference RawVariable in a generic method (line 48)?
 6. In GetInstance, you seems to always return a new instance, meaning calling GetIdentityReader will always execute the reflection, I think.
 7. Converting all the string to then use in comparison/grouping is possible, but not so nice.

I summary, check this with the Dev team and they might already have a solution for this. If now, they can help you to write this better.

tkukuljan: Thanks! I proceeded with contacting the dev team.
Line 48 is a mistake, the method GetInstance works well. 
I did not find a better way of grouping than with a string. I wouldn't mind a suggestion :)

ifrs17/Import/Importers.ipynb
amuolo: this comment is outdated and not relevant anymore because new version of the method with improvements has been introduced

ifrs17/Import/Importers.ipynb
amuolo: this is also outdated, resolving it

ifrs17/Utils/EqualityComparers.ipynb
amuolo: this is done

ifrs17/Utils/EqualityComparers.ipynb
amuolo: outdated, resolving

ifrs17/Utils/IdentityPropertyReader.ipynb
tkukuljan: The Dev team indeed gave a better way (ToString). See the new version.

amuolo (approved):

ifrs17/Import/Importers.ipynb
```diff
@@ -541,6 +519,100 @@
...
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public async static Task ValidateForMandatoryAocSteps(this IWorkspace workspace, IDataSet dataSet, HashSet<AocStep> mandatoryAocSteps)",
+                "\n{",
+                "\n    var includedProperties = typeof(RawVariable).GetIdentityProperties().Select(x=>x.Name).Except(new[]{nameof(AocType), nameof(Novelty)});",
+                "\n    var missingAocStepsByIdentityProperties = (await workspace.Query<RawVariable>().ToListAsync())",
+                "\n                     .GroupBy(x => string.Join(\", \", x.ToString().Split('{','}')[1].Split(',').Where(y=> includedProperties.Contains(y.Split('=')[0].Trim()))).Trim(),",
```
dcolleoni, line 604: what is this line doing? can we simplify it?

tkukuljan: -the following. if i set:
var x = new RawVariable{DataNode = "A", EstimateType = "B"};
-then
x.ToString()
-prints: 
RawVariable { DataNode = A, AocType = , Novelty = , Id = 00000000-0000-0000-0000-000000000000, Partition = 00000000-0000-0000-0000-000000000000, AmountType = , AccidentYear = , Values = , EstimateType = B }
-then the long line takes:
DataNode = A, AmountType = , AccidentYear = , EstimateType = B

AlekseiNekrasov: Can't just use x.Select() instead of the splits?

AlekseiNekrasov: image

tkukuljan: Select works if one knows all the properties they want to select. The goal of the method was to work also if the data model changes. Then I did not know how to make it work with the Select (or equivalent in the GroupBy clause).

ifrs17/Import/Importers.ipynb
```diff
@@ -541,6 +519,100 @@
...
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public async static Task ValidateForMandatoryAocSteps(this IWorkspace workspace, IDataSet dataSet, HashSet<AocStep> mandatoryAocSteps)",
+                "\n{",
+                "\n    var includedProperties = typeof(RawVariable).GetIdentityProperties().Select(x=>x.Name).Except(new[]{nameof(AocType), nameof(Novelty)});",
```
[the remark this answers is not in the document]
tkukuljan: The reflection gets executed once per Cashflow import, which is much less costly than reading the entire file.
Therefore I think the change does not improve the efficiency enough to be worth implementing it.

ifrs17/Import/Importers.ipynb
dcolleoni: there is a ToIdentityString() method. What about this one?

tkukuljan: Good proposition.
I included it in the new version.

ifrs17/Import/Importers.ipynb
```diff
@@ -593,6 +571,100 @@
...
+                "\n"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "markdown",
+            "source": [
+                "## Validate for Data Node States Logic"
```
dcolleoni, line 610: This is repeated. I guess come from the merge conflict resolution. Have a look overall at the file to spot any other mistake like this one

ifrs17/Import/Importers.ipynb
```diff
@@ -593,6 +571,100 @@
...
+                "public async static Task ValidateForMandatoryAocSteps(this IWorkspace workspace, IDataSet dataSet, HashSet<AocStep> mandatoryAocSteps)",
+                "\n{",
+                "\n    var excludedProperties = new[]{nameof(AocType), nameof(Novelty)};",
+                "\n    var missingAocStepsByIdentityProperties = (await workspace.Query<RawVariable>().ToListAsync())",
+                "\n                     .GroupBy(x => string.Join(\", \", x.ToIdentityString().Split(',').Where(y=> y != \" \" && !excludedProperties.Contains(y.Split(':')[0]))),",
+                "\n                              x => new AocStep(x.AocType, x.Novelty),",
+                "\n                              (properties, parsedAocSteps) => (properties, mandatoryAocSteps.Except(parsedAocSteps))",
+                "\n                             );",
+                "\n    foreach((var properties, var missingSteps) in missingAocStepsByIdentityProperties) ",
+                "\n        foreach(var missingStep in missingSteps)  ApplicationMessage.Log(Warning.MandatoryAocStepMissing, missingStep.AocType, missingStep.Novelty, properties);",
```
dcolleoni, line 661: I would log an Error. Why a Warning?

ifrs17/Import/Importers.ipynb
```diff
@@ -593,6 +571,100 @@
...
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public async static Task ValidateForMandatoryAocSteps(this IWorkspace workspace, IDataSet dataSet, HashSet<AocStep> mandatoryAocSteps)",
+                "\n{",
+                "\n    var excludedProperties = new[]{nameof(AocType), nameof(Novelty)};",
+                "\n    var missingAocStepsByIdentityProperties = (await workspace.Query<RawVariable>().ToListAsync())",
+                "\n                     .GroupBy(x => string.Join(\", \", x.ToIdentityString().Split(',').Where(y=> y != \" \" && !excludedProperties.Contains(y.Split(':')[0]))),",
```
dcolleoni, line 656: Let's allow ToIdentityString() to take inputs []params => dimension names of property we want to omit from the returned string. That should make this lambda look like: 
x => x.ToIdentityString(nameof(AocType),nameof(Novelty))

ifrs17/Import/Importers.ipynb
tkukuljan: This was chosen so by Daniel, with a reason that I don't know.

ifrs17/Import/Importers.ipynb
tkukuljan: Should I talk to devs about it?
In that case it will take I while I think.

ifrs17/Import/Importers.ipynb
dcolleoni: ok, let's leave it for the time being.

ifrs17/Import/Importers.ipynb
tkukuljan: Did in the newest commit

dcolleoni (approved):

PR #154, opened 2022-12-23
Yield curve Name into Queries

scolella:
PUBLISHED


ifrs17/Utils/Queries.ipynb
```diff
@@ -217,14 +217,27 @@
...
-                "\n    return (await querySource.LoadCurrentAndPreviousParameterAsync<YieldCurve>(args, ",
+                "public static async Task<Dictionary<string, Dictionary<int, YieldCurve>>> LoadCurrentYieldCurveAsync(this IQuerySource querySource, Args args,",
+                "\n                                                                                                                                IEnumerable<DataNodeData> dataNodes)",
+                "\n{    ",
+                "\n    var currentYieldCurveByGoc = new Dictionary<string,  Dictionary<int, YieldCurve>>();",
+                "\n    ",
+                "\n    foreach (var dn in dataNodes)",
+                "\n    {",
+                "\n        var argsNew = args with {Year = dn.Year, Month = MonthInAYear, Scenario = args.Scenario};",
+                "\n        var loadedYc = await querySource.LoadCurrentAndPreviousParameterAsync<YieldCurve>(argsNew, ",
```
dcolleoni, line 228: args should be used instead of argsNew (remove)

scolella: removed argsNew

ifrs17/Utils/Queries.ipynb
```diff
@@ -183,18 +183,18 @@
...
                 "\n    var lockedInYieldCurveByGoc = new Dictionary<string, YieldCurve>();",
                 "\n    foreach (var dn in dataNodes.Where(x => x.ValuationApproach == ValuationApproaches.BBA))",
                 "\n    {",
                 "\n        var argsNew = args with {Year = dn.Year, Month = dn.Month, Scenario = args.Scenario};",
-                "\n        var loadedYc = (await querySource.LoadCurrentParameterAsync<YieldCurve>(argsNew, x => x.Currency, x => x.Currency == dn.ContractualCurrency));",
-                "\n        ",
-                "\n        if(!loadedYc.TryGetValue(dn.ContractualCurrency, out var lockedYc))",
+                "\n        var loadedYc = (await querySource.LoadCurrentParameterAsync<YieldCurve>(argsNew, x => x.Currency, x => x.Currency == dn.ContractualCurrency && x.Name == dn.YieldCurveName));",
+                "\n                                    ",
+                "\n        if (!loadedYc.TryGetValue(dn.ContractualCurrency, out var lockedYc) || lockedYc == null)",
```
dtrzesniak, line 195: How can this happen? If on the dictionary, the bool is false anyway. Otherwise, lockedYc should never be null.

ifrs17/Utils/Queries.ipynb
```diff
@@ -217,14 +217,26 @@
...
                 "\n                                                                               x => x.Currency,",
-                "\n                                                                               x => contractualCurrenciesInScope.Contains(x.Currency)",
-                "\n                                                                               ));",
+                "\n                                                                               x => x.Currency == dn.ContractualCurrency ",
+                "\n                                                                                    && (dn.ValuationApproach == ValuationApproaches.VFA ",
+                "\n                                                                                                                ? x.Name == dn.YieldCurveName ",
+                "\n                                                                                                                : true));",
+                "\n        if(!loadedYc.TryGetValue(dn.ContractualCurrency, out var currentYcDict) || currentYcDict == null || currentYcDict?.Values == null)",
```
dtrzesniak, line 233: These things should never happen --> any check on YC should be in the setter, not the getter.

ifrs17/Utils/Queries.ipynb
```diff
@@ -217,14 +217,26 @@
...
+                "\n    var currentYieldCurveByGoc = new Dictionary<string,  Dictionary<int, YieldCurve>>();",
+                "\n    ",
+                "\n    foreach (var dn in dataNodes)",
+                "\n    {",
+                "\n        var loadedYc = await querySource.LoadCurrentAndPreviousParameterAsync<YieldCurve>(args, ",
                 "\n                                                                               x => x.Currency,",
-                "\n                                                                               x => contractualCurrenciesInScope.Contains(x.Currency)",
-                "\n                                                                               ));",
+                "\n                                                                               x => x.Currency == dn.ContractualCurrency ",
+                "\n                                                                                    && (dn.ValuationApproach == ValuationApproaches.VFA ",
```
dtrzesniak, line 230: I don't understand this part. The current curve is always used, why this?

ifrs17/Utils/Queries.ipynb
```diff
@@ -217,14 +217,26 @@
...
-                "\n                                                                                                      IEnumerable<DataNodeData> dataNodes)",
-                "\n{",
-                "\n    var contractualCurrenciesInScope = dataNodes.Select(dn => dn.ContractualCurrency).ToHashSet();",
-                "\n    return (await querySource.LoadCurrentAndPreviousParameterAsync<YieldCurve>(args, ",
+                "public static async Task<Dictionary<string, Dictionary<int, YieldCurve>>> LoadCurrentYieldCurveAsync(this IQuerySource querySource, Args args,",
+                "\n                                                                                                                                IEnumerable<DataNodeData> dataNodes)",
+                "\n{    ",
+                "\n    var currentYieldCurveByGoc = new Dictionary<string,  Dictionary<int, YieldCurve>>();",
+                "\n    ",
+                "\n    foreach (var dn in dataNodes)",
```
dtrzesniak, line 225: The current curve is not DN dependent, right? Why do you need this at all?

ifrs17/Utils/Queries.ipynb
AlekseiNekrasov: @dtrzesniak Daniel, do you mean checking additional "|| lockedYc == null" ? That one makes no sense since TryGetValue controls enough.

dtrzesniak: Exactly... and there should never be a null value in this dictionary

ifrs17/Utils/Queries.ipynb
AlekseiNekrasov: Should that check be done inside the task "LoadCurrentAndPreviousParameterAsync" ?

image

ifrs17/Test/QueriesTest.ipynb
```diff
@@ -528,7 +528,7 @@
             "cell_type": "code",
             "source": [
                 "public async Task<ActivityLog> CheckLoadYieldCurveAsync(Args args, IEnumerable<YieldCurve> testData, ",
-                "\n                                           (int year , int month) dataNodeYearAndMonth, ",
+                "\n                                           (int year , int month, string yieldCurveName, string valuationApproach) dataNodeParameters, ",
```
dcolleoni, line 531: rename : parameters have already a meaning and are something else. Used dataNodeTestData

ifrs17/Test/QueriesTest.ipynb
```diff
@@ -590,6 +597,84 @@
...
+            "source": [
+                "var args = new Args(\"CH\",2020,9,Periodicity.Monthly,null);",
+                "\nvar testData = new YieldCurve[] {new YieldCurve{ Currency = \"EUR\", Year = 2020, Month = 6, Name = \"B\", Values = new double[]{0.1,0.2,0.3,0.4,0.5,0.6} }, ",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2020, Month = 6, Name = \"B\", Values = new double[]{2.1,2.2,2.3,2.4,2.5,2.6} }, ",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2016, Month = 3, Name = \"C\", Values = new double[]{4.1,4.2,4.3,4.4,4.5,4.6} },",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2016, Month = 3, Name = \"B\", Values = new double[]{6.1,6.2,6.3,6.4,6.5,6.6} }};",
+                "\n",
+                "\n(int year , int month, string yieldCurveName, string valuationApproach) dataNodeParameters = (2016, 2, \"A\", \"BBA\");",
+                "\n",
+                "\nvar activity = await CheckLoadYieldCurveAsync(args, testData, dataNodeParameters, 6.1, 0.1, 6.1);"
```
dcolleoni, line 662: don't enter a BM if you expect failure. use **default** instead

ifrs17/Test/QueriesTest.ipynb
```diff
@@ -590,6 +597,84 @@
...
+        {
+            "cell_type": "code",
+            "source": [
+                "var args = new Args(\"CH\",2020,9,Periodicity.Monthly,null);",
+                "\nvar testData = new YieldCurve[] {new YieldCurve{ Currency = \"EUR\", Year = 2020, Month = 6, Values = new double[]{0.1,0.2,0.3,0.4,0.5,0.6} }, ",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2020, Month = 3, Values = new double[]{2.1,2.2,2.3,2.4,2.5,2.6} }, ",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2021, Month = 9, Values = new double[]{4.1,4.2,4.3,4.4,4.5,4.6} },",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2016, Month = 3, Values = new double[]{6.1,6.2,6.3,6.4,6.5,6.6} }};",
+                "\n",
+                "\n(int year , int month, string yieldCurveName, string valuationApproach) dataNodeParameters = (2016, 6, null, \"VFA\");",
```
dcolleoni, line 634: Properly test the VFA. 
Now YieldCurveName is not used nor in the YC neither in the DataNode

ifrs17/Utils/Queries.ipynb
AlekseiNekrasov: @dtrzesniak Fixed with the last commit.

ifrs17/Utils/Queries.ipynb
AlekseiNekrasov: @dtrzesniak Fixed with the last commit.

ifrs17/Test/QueriesTest.ipynb
```diff
@@ -590,6 +597,84 @@
             "execution_count": 0,
             "outputs": []
         },
+        {
+            "cell_type": "code",
+            "source": [
+                "var args = new Args(\"CH\",2020,9,Periodicity.Monthly,null);",
+                "\nvar testData = new YieldCurve[] {new YieldCurve{ Currency = \"EUR\", Year = 2020, Month = 6, Name = \"A\", Values = new double[]{0.1,0.2,0.3,0.4,0.5,0.6} }, ",
```
dcolleoni, line 604: make this case more interesting by adding some YC with Name = null
a Name A within 2016 and 2020 to update CurrentPrevious result

ifrs17/Test/QueriesTest.ipynb
AlekseiNekrasov: Renamed dataNodeParameters to dataNodeTestData.

ifrs17/Test/QueriesTest.ipynb
AlekseiNekrasov: Fixed.

ifrs17/Test/QueriesTest.ipynb
scolella: done

ifrs17/Test/QueriesTest.ipynb
scolella: done

scolella: done

ifrs17/Utils/Queries.ipynb
AlekseiNekrasov: Sara: The current curve is dn dependendent beacuse of YieldCurveName

ifrs17/Utils/Queries.ipynb
AlekseiNekrasov: Sara: It is always used, but the additional filter in the query for YieldCurveName happens only for VFA

ifrs17/Utils/Queries.ipynb
dtrzesniak: But then you should extract the name, no? You are calling the query for every DN, even if the name is null for all. The foreach should be on the distinct names + the null.

ifrs17/Utils/Queries.ipynb
dtrzesniak: In connection with the name above, I might now understand.
I don't see why it is related to VFA though... if there is a name in the dn, you should take the curve with the name, right?
Basically you are saying to ignore the name if the valuation approach is not VFA.
Since we don't have VFA, I would not add this logic, we can think later.
I remember talking to @dcolleoni about this case, that I think says for the current curve always ignore the name.

ifrs17/Utils/Queries.ipynb
```diff
@@ -183,18 +183,18 @@
...
             "source": [
-                "public static async Task<Dictionary<string, YieldCurve>> LoadLockedInYieldCurveAsync(this IQuerySource querySource, Args args, ",
-                "\n                                                                                                       IEnumerable<DataNodeData> dataNodes)",
+                "public static async Task<Dictionary<string, YieldCurve>> LoadLockedInYieldCurveAsync(this IQuerySource querySource, Args args, IEnumerable<DataNodeData> dataNodes)",
+                "\n",
                 "\n{",
                 "\n    var lockedInYieldCurveByGoc = new Dictionary<string, YieldCurve>();",
                 "\n    foreach (var dn in dataNodes.Where(x => x.ValuationApproach == ValuationApproaches.BBA))",
                 "\n    {",
                 "\n        var argsNew = args with {Year = dn.Year, Month = dn.Month, Scenario = args.Scenario};",
```
dtrzesniak, line 192: BTW, you can remove this, it's not needed.

ifrs17/Utils/Queries.ipynb
```diff
@@ -183,18 +183,18 @@
         {
             "cell_type": "code",
             "source": [
-                "public static async Task<Dictionary<string, YieldCurve>> LoadLockedInYieldCurveAsync(this IQuerySource querySource, Args args, ",
-                "\n                                                                                                       IEnumerable<DataNodeData> dataNodes)",
+                "public static async Task<Dictionary<string, YieldCurve>> LoadLockedInYieldCurveAsync(this IQuerySource querySource, Args args, IEnumerable<DataNodeData> dataNodes)",
+                "\n",
                 "\n{",
                 "\n    var lockedInYieldCurveByGoc = new Dictionary<string, YieldCurve>();",
                 "\n    foreach (var dn in dataNodes.Where(x => x.ValuationApproach == ValuationApproaches.BBA))",
```
dtrzesniak, line 190: I think also this needs improvement.
We don't need this query so many times.

ifrs17/Utils/Queries.ipynb
dcolleoni: I would leave it now as it is done already...

ifrs17/Utils/Queries.ipynb
dcolleoni: I understand what you mean. And I think it should work. One has that to adapt the getter in the storage accordingly. 
If we put the YCName as key of the dict we can define and use a defaultKey = "__D" in const.

ifrs17/Utils/Queries.ipynb
scolella: for locked we need to update the args for each dn. Right @dcolleoni ?

ifrs17/Utils/Queries.ipynb
scolella: If we need to update the args, then we can not optimize it, I think

scolella: the optimization of the locked query will be done in a following PR

ifrs17/Utils/Queries.ipynb
scolella: the optimization of the locked query will be done in a following PR

ifrs17/Utils/Queries.ipynb
scolella: implemented

dcolleoni (approved):

ifrs17/Utils/Queries.ipynb
dcolleoni, line 192: shouldn't this be monthInAYear instead of dn.Month?

ifrs17/Test/QueriesTest.ipynb
```diff
@@ -570,11 +576,99 @@
...
+                "\n                                 new YieldCurve{ Currency = \"CHF\", Year = 2020, Month = 6, Name = \"A\", Values = new double[]{3.1,3.2,3.3,3.4,3.5,3.6} }, ",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2020, Month = 6, Name = \"B\", Values = new double[]{2.1,2.2,2.3,2.4,2.5,2.6} }, ",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2016, Month = 3, Name = \"A\", Values = new double[]{6.1,6.2,6.3,6.4,6.5,6.6} },",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2016, Month = 3, Values = new double[]{9.1,9.2,9.3,9.4,9.5,9.6} },",
+                "\n                                 new YieldCurve{ Currency = \"CHF\", Year = 2016, Month = 3, Name = \"A\", Values = new double[]{8.1,8.2,8.3,8.4,8.5,8.6} },",
+                "\n                                 new YieldCurve{ Currency = \"EUR\", Year = 2016, Month = 3, Name = \"B\", Values = new double[]{7.1,7.2,7.3,7.4,7.5,7.6} }};",
```
dcolleoni, line 613: add here 2016,12,"A"
this should be the locked in curve used

scolella: I changed accordingly the following test. t is more intuitive in this way

ifrs17/Utils/Queries.ipynb
scolella: done

PR #161, opened 2023-01-06
Yield curve import triggers calculations

calderinisystemorph:


ifrs17/Constants/Validations.ipynb
```diff
@@ -157,7 +157,7 @@
                 "\n    (Error.CreditDefaultRateNotFound      , 1) => $\"Credit Default Rate not found for rating {s[0]}.\",",
                 "\n    (Error.MissingPremiumAllocation       , 1) => $\"Premium Allocation Rate not found for Group of Contract {s[0]}.\",  // TODO: this is now a warning to be produced by a validation in the importers (default is 1)",
                 "\n    (Error.ReinsuranceCoverage            , 1) => $\"Reinsurance Allocation Rate not found for Group of Insurance Contract {s[0]}.\",",
-                "\n    (Error.YieldCurveNotFound             , 3) => $\"Yield Curve not found for currency {s[0]}, year {s[1]}, and month {s[2]}.\",",
+                "\n    (Error.YieldCurveNotFound             , 4) => $\"Yield Curve not found for Currency {s[0]}, Year {s[1]}, Month {s[2]}, and Scenario {(s[3] == null ? \"Best Estimate\" : s[3])}.\",",
```
dtrzesniak, line 160: Also Name, as part of the YC model?

ifrs17/Constants/Validations.ipynb
```diff
@@ -183,7 +183,8 @@
                 "public static string Get (Warning w, params string[] s) => (w, s.Length) switch {",
                 "\n    // Import",
                 "\n    (Warning.ActiveDataNodeWithCashflowBOPI , 1) => $\"Cash flow with AoC Type: {AocTypes.BOP} and Novelty: {Novelties.I} for Group of Contract {s[0]} is not allowed because previous period data are available.\",",
-                "\n    (Warning.ScenarioReCalculations         , 1) => $\"The import of the current file for the Best Estimate scenario makes the result of dependent Scenarios out of date. Hence, the following Scenarios are recalculated: {s[0]}.\", ",
+                "\n    (Warning.VariablesAlreadyImported       , 0) => $\"The import of the current file does not contain any new data. Hence, no data will be saved or calculations performed\",",
```
dtrzesniak, line 186: English is weird: I would add "will be" before performed

ifrs17/Import/ImportStorage.ipynb
```diff
@@ -145,7 +145,7 @@
                 "\n        await hierarchyCache.InitializeAsync<AmountType>();",
                 "\n        ",
                 "\n        //EstimateType to load and to update",
-                "\n        EstimateTypesByImportFormat = new InputSource[] { InputSource.Opening, InputSource.Actual, InputSource.Cashflow,}",
+                "\n        EstimateTypesByImportFormat = new InputSource[] { InputSource.Opening, InputSource.Actual, InputSource.Cashflow, }",
```
dtrzesniak, line 148: extra comma

ifrs17-template/Test/ScenarioYieldCurveImportTest.ipynb
```diff
@@ -0,0 +1,529 @@
...
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "diffs.Count().Should().NotBe(0);"
+            ],
+            "metadata": {},
```
dtrzesniak, line 497: Many empty cells?

ifrs17/Import/Importers.ipynb
```diff
@@ -357,9 +355,57 @@
...
+                "\nThese are the methods used to create the partition if not already existing in the DataSource. "
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public async Task SetPartitionAsync<IPartition>(ImportArgs args, params IDataSource[] dataSources)",
```
dtrzesniak, line 369: I don't get this name. SetPartitionAsync, bit you are updating/committing, right?

ifrs17/Import/Importers.ipynb
```diff
@@ -373,32 +419,7 @@
                 "\n    var args = GetArgsFromMain(dataSet);",
                 "\n    if(ApplicationMessage.HasErrors()) return null;",
```
dtrzesniak, line 420: Why this check here? This should never be null at this point. If it is, we miss a validation earlier in the process.

ifrs17-template/Test/scenarioYieldCurve.csv
```diff
@@ -0,0 +1,9 @@
+@@Main
```
dcolleoni, line 1: for the file name, use the same convention we have for the other files:
YieldCurve_2020_12_YCUP1.csv

ifrs17/Constants/Validations.ipynb
dcolleoni: end the message with a full stop.

ifrs17/Import/Importers.ipynb
```diff
@@ -675,42 +694,74 @@
...
+                "\n    var toCommitYieldCurves  = await workspace.Query<YieldCurve>().ToArrayAsync();",
+                "\n    if (!toCommitYieldCurves.Any()) {",
                 "\n        ApplicationMessage.Log(Warning.VariablesAlreadyImported);   ",
                 "\n        return Activity.Finish().Merge(importLog);",
                 "\n    }",
-                "\n    ",
-                "\n    await DataSource.UpdateAsync<YieldCurve>(toCommitYieldCurves);",
-                "\n    await DataSource.CommitAsync();",
+                "\n",
+                "\n    PartitionByReportingNodeAndPeriod[] targetPartitions;",
```
dcolleoni, line 724: I would try to make this method easier to read by extracting some methods. 
For example these lines to set the targetPartitions could be a method. 
Moreover, other importers require a similar functionality. For example Cashflow format uses a function call getAllArgs which (I think) serves a similar purpose. I suggest to unify the methods (possibly avoiding some code repetition). It is possible this methos would require the importFormat as an input to discriminate specific logic, but I think this is a cost we can pay to improve readability.

dcolleoni: To increase readability I would also use the same variable names across different import format. 
eg in cashflow importer we loop over allArgs, here we loop over targetPartitions, but the meaning is exactly the same.

ifrs17/Import/Importers.ipynb
dcolleoni: I think it is important we transparently name the method that commit to datasource

ifrs17/Import/Importers.ipynb
```diff
@@ -487,16 +509,13 @@
             "source": [
                 "public async Task<ImportArgs[]> GetAllArgsAsync(ImportArgs primaryArgs)",
                 "\n{",
-                "\n    if (primaryArgs.Scenario != null)",
-                "\n        return primaryArgs.RepeatOnce().ToArray();",
+                "\n    if (primaryArgs.Scenario != null) return primaryArgs.RepeatOnce().ToArray();",
```
dcolleoni, line 512: This is a style change. I personally prefer going to new line and I always do that. 
Let's align on this.

ifrs17/Constants/Validations.ipynb
amuolo: done

ifrs17/Constants/Validations.ipynb
amuolo: done

ifrs17/Import/ImportStorage.ipynb
amuolo: corrected

ifrs17-template/Test/ScenarioYieldCurveImportTest.ipynb
amuolo: deleted

ifrs17/Import/Importers.ipynb
amuolo: "Set" is a bit confusing because this method is not setting the partition.  
What do you propose? Does "UploadPartitionAsync" read better? 
@dtrzesniak @dcolleoni

ifrs17/Import/Importers.ipynb
amuolo: We have validations in the GetArgsFromMain method. The check in line 374 serves to check whether those validations failed, and in case they did, it has to return.

ifrs17-template/Test/scenarioYieldCurve.csv
amuolo: done

ifrs17/Import/Importers.ipynb
amuolo: done

ifrs17/Import/Importers.ipynb
amuolo: done! GetAllArgs is now used for both the transactional data and the yield curve importers. Also, this means that the variable names are now consistent.

ifrs17/Import/Importers.ipynb
dtrzesniak: Yes, something along these lines... Set is confusing, as you say, as it is not setting.
Upload is ok, as I think we also use this name in the same context as here.

ifrs17/Import/Importers.ipynb
dtrzesniak: I understand, but shouldn't we do this inside the GetArgsFromMain, imposing that this method always gives valid args?
Also, if you use GetArgsFromMain in other places, you would have to implement similar checks. Thus IMO the check must be within the method, you see?

dtrzesniak: I am talking on like 375, not 374.

amuolo: GetArgsFromMain can return an args with ReportingNode null. This happens for instance in the YieldCurve importer, for which the main tab has only Year Mont Scenario. Conversely, GetArgsFromMain has other kind of checks, e.g. main tab exists, has entry, etc. Therefore, since the check in line 375 is specific for the GetArgsAndCommitPartitions method, because the intention is also to commit the partition and hence the ReportingNode must be provided, this line was placed (not by this PR) outside of the method GetArgsFromMain.

ifrs17/Import/Importers.ipynb
amuolo: Ok, and what about CommitPartitionAsync? I think this is the best name because it is consistent with the method GetArgsAndCommitPartitionAsync. How do you like it?

dtrzesniak: It's fine with me.

dtrzesniak (approved):

dcolleoni:
Can we add a Toggle or any other switch to switch off the scenario functionality especially around the recalculation of the YieldCurve?

ifrs17/Import/Importers.ipynb
amuolo: Let's discuss and have a dedicated ticket.

ifrs17/Import/Importers.ipynb
```diff
@@ -485,19 +507,42 @@
...
+                "\n        {",
+                "\n            if(args.Scenario == null) {",
+                "\n                var scenariosWithYieldCurves = await dataSource.Query<YieldCurve>().Where(x => x.Year == args.Year && x.Month == args.Month && x.Scenario != null)",
+                "\n                    .Select(x => x.Scenario).Distinct().ToArrayAsync();",
+                "\n                var targetPartitions = await dataSource.Query<PartitionByReportingNodeAndPeriod>()",
+                "\n                    .Where(x => x.Year == args.Year && x.Month == args.Month && !scenariosWithYieldCurves.Contains(x.Scenario)).OrderBy(x => x.Scenario).ToArrayAsync();",
+                "\n                var targetScenarios = targetPartitions.Where(x => x.Scenario != null).Select(x => x.Scenario);",
+                "\n                if(targetScenarios.Any()) ApplicationMessage.Log(Warning.ScenarioReCalculations, String.Join(\", \", targetScenarios));",
+                "\n                allArgs = targetPartitions.Select(x => new ImportArgs(x.ReportingNode, x.Year, x.Month, default(Periodicity), x.Scenario, ImportFormats.Cashflow)).ToArray();",
```
dcolleoni, line 524: don't we need to add args here?

ifrs17/Import/Importers.ipynb
```diff
@@ -485,19 +507,42 @@
...
+                "\n            }",
+                "\n            else {",
+                "\n                allArgs = (await dataSource.Query<PartitionByReportingNodeAndPeriod>()",
+                "\n                    .Where(x => x.Year == args.Year && x.Month == args.Month && x.Scenario == null).ToArrayAsync())",
```
dcolleoni, line 528: Why do we filter for Scenario null?
I understand this is the case of an import of YC for a scenario. Then I don't expect any recalculation for BE.

ifrs17/Import/Importers.ipynb
```diff
@@ -673,44 +718,64 @@
...
+                "\n        var yieldCurve = new YieldCurve { ",
+                "\n            Currency = datarow.Field<string>(nameof(YieldCurve.Currency)),",
+                "\n            Year = primaryArgs.Year,",
+                "\n            Month = primaryArgs.Month, ",
+                "\n            Scenario = primaryArgs.Scenario,",
+                "\n            Values = datarow.Table.Columns.Where(c => c.ColumnName.StartsWith(nameof(YieldCurve.Values))).OrderBy(c => c.ColumnName.Length).ThenBy(c => c.ColumnName)",
+                "\n                            .Select(x => datarow.Field<string>(x.ColumnName).CheckStringForExponentialAndConvertToDouble()).ToArray(),",
+                "\n            Name = hasNameColumn ? datarow.Field<string>(nameof(YieldCurve.Name)) : default(string)",
+                "\n        };",
+                "\n        return committedYieldCurves.Contains(yieldCurve, YieldCurveComparer.Instance()) ? null : yieldCurve;",
```
dcolleoni, line 740: I think this yield curve comparer needs improvement. Do we have an issue for that? @calderinisystemorph 
For example importing the same YC for 2 different months is not required and then a new YC should be stored. 
@amuolo can you add this issue to the board? thanks

ifrs17/Import/Importers.ipynb
```diff
@@ -673,44 +718,64 @@
...
+                "\n        if(ApplicationMessage.HasErrors()) return Activity.Finish().Merge(importLog);",
+                "\n        ",
+                "\n        // Avoid starting the computation if no best estimate cash flow has ever been imported ",
+                "\n        if(args.Scenario == null) {",
+                "\n            await options.TargetDataSource.Partition.SetAsync<PartitionByReportingNodeAndPeriod>(null);",
+                "\n            if(!(await options.TargetDataSource.Query<RawVariable>().Where(x => x.Partition == targetPartition).Take(1).ToArrayAsync()).Any()) continue;",
+                "\n        }",
+                "\n",
+                "\n        // Remove data nodes which are unaffected by the updated yield curves",
+                "\n        await workspaceToCompute.DeleteAsync( await workspaceToCompute.Query<RawVariable>()",
```
dcolleoni, line 770: what is the type that is deleted here? is it RawVariable? why?

ifrs17/Import/Importers.ipynb
amuolo: No because for this case the args (taken from the Main tab) is not usable. It doesn't contain the ReportingNode and only serves to calculate all relevant args to be computed.

ifrs17/Import/Importers.ipynb
amuolo: Yes this is the case of an import of scenario YC. In fact there is no recalculation for BE. However, we need a list of all BE partitions for the target period in order to start calculations for all the resulting reporting nodes.

ifrs17/Import/Importers.ipynb
amuolo: The yield curve importer triggers calculation with ImportFormat Cashflow. Therefore, the parsedRawVar and parsedIfrsVar in the ImportStorage will be full and empty respectively. However, the former can be pruned in order to select only those raw variables connected to data nodes having the same currency as one of those newly imported. This is why we have this step here where we delete useless raw variables.

ifrs17/Import/Importers.ipynb
amuolo: done #191

dcolleoni (approved):

PR #163, opened 2023-01-10
Parameters report

dtrzesniak:
This should be the 1st step of the solution.


ifrs17-template/Report/ParameterReports.ipynb
```diff
@@ -0,0 +1,521 @@
...
+                "\n    .ToArray();"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "var dataNodeStates = (await Workspace.LoadCurrentAndPreviousParameterAsync<DataNodeState>(args, x => x.DataNode))",
```
dcolleoni, line 225: shouldn't we filter here for the current period?

ifrs17-template/Report/ParameterReports.ipynb
```diff
@@ -0,0 +1,521 @@
...
+                "\n        GroupOfContract = x.Key,",
+                "\n        YieldCurveType = \"Locked-In Curve\",",
+                "\n        Year = x.Value.Year,",
+                "\n        Month = x.Value.Month,",
+                "\n        Scenario = x.Value.Scenario,",
+                "\n        Currency = x.Value.Currency,",
+                "\n        Name = x.Value.Name }",
+                "\n        )",
+                "\n    .ToArray();",
+                "\nvar currentYieldCurves = (await Workspace.LoadCurrentYieldCurveAsync(args, dataNodeData))",
```
dcolleoni, line 266: There's no property to distinguish between Current YC for period -1 and period 0. It will be noticeable only from Year Month if 2 different curves are taken, else also Year and Month are the same. 
Should we use the YieldCurveName to specify "for current period" - "for previous period"?

ifrs17-template/Report/ParameterReports.ipynb
```diff
@@ -0,0 +1,521 @@
...
+        {
+            "cell_type": "code",
+            "source": [
+                "await Report.ForObjects(partnerDefaultRates)",
+                "\n            .WithQuerySource(Workspace)",
+                "\n            .GroupRowsBy(x => x.CreditRiskRating)",
+                "\n            .GroupColumnsBy(x => x.Period)",
+                "\n            .ToTable()",
+                "\n            .ExecuteAsync()"
+            ],
```
dcolleoni, line 515: I would separate this NB in 2. One with dataModel and queries, one with reports.

dtrzesniak: Being done

ifrs17/Report/ParameterReportsQueries.ipynb
```diff
@@ -0,0 +1,400 @@
...
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public static async Task<DataNodeStateReportParameter[]> GetDataNodeStateReportParameters(this IWorkspace workspace, ImportArgs args) =>",
+                "\n    (await workspace.LoadCurrentAndPreviousParameterAsync<DataNodeState>(args, x => x.DataNode))",
```
dcolleoni, line 192: I was checking this query I don't see where args.ReportingNode is used in the filter. I think the query expects the workspace to have a  partition set. 
We can improve this in a following PR.

ifrs17-template/Report/ParameterReports.ipynb
```diff
@@ -0,0 +1,252 @@
...
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "await Report.ForObjects(partnerDefaultRates)",
```
dcolleoni, line 240: Ideally we report ForDataCube. We can take it in the nect PR

dcolleoni (approved):

PR #176, opened 2023-01-16
Projections

scolella:
Projection


ifrs17/DataModel/DataStructure.ipynb
```diff
@@ -1783,6 +1783,8 @@
                 "public record IfrsVariable : BaseDataRecord",
                 "\n{",
                 "\n    public double Value { get; init; }",
+                "\n    public double Value1 { get; init; }",
```
dtrzesniak, line 1786: I was wondering why we did this here, also in V7. The RawVar have double[], the Report Variable have Value and Projection as properties. Isn't there a way we can do that without hardcoding the number of projections?

dcolleoni:
I am closing this PR as duplicate of #197 
@AlekseiNekrasov  pls take note of @dtrzesniak comment

Issue #167, opened 2023-01-13
Improve Activity log methods


dcolleoni:
Improvements on how the logs are stored in the DB:

andrey-katz-systemorph:
Suggestions for improvement: 

1. If the import or export fail, they should not have a "fake" source Id. This should be null 
2. Change the API as suggested by @dtrzesniak : WithStoreActivityLog().ExecuteAsync()

dcolleoni:
Further improvements on the code taken from comments in PR #178 

Can't you simplify this:

Add the "await DataSource.UpdateAsync(importFile.RepeatOnce());"," inside the try
Remove importSucceeded,
Add "activity = activity with {Category = "Import from File"};"," inside the try and catch --> you are creating activity twice (once to set SourceId, once to set Category).
Declare importFile inside the try
Extract the try/catch as method of type and pass the parameters to avoid the repetition of this code.
I have not thought it through, but I guess this should work and make the code leaner

PR #178, opened 2023-01-17
167 activity log v2

andrey-katz-systemorph:
Implement a slightly different API for saving the logs
.WithLogsRecordedInDb(....).ExecuteAsync() instead of ExecuteWithStoreActivityAsync() and changes all the PV and template uses accordingly. Changing the type of SourceId from Guid to Guid? to avoid fake IDs in failed imports.


ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -354,6 +354,160 @@
...
+                "\n}"
+            ],
+            "metadata": {},
+            "execution_count": 0,
+            "outputs": []
+        },
+        {
+            "cell_type": "code",
+            "source": [
+                "public static ImportBuilderWriter WithLogsRecordedInDb(this ImportOptionsBuilder builder, ",
```
dtrzesniak, line 479: It just bothers me the InDb in the namehere, as this would also work if you don't have a Db (say, in memory --> you would also create a log in the datasource).

What about just .WithLogRecords? You are anyway passing IDataSource as parameter, and this should make clear where the log records are being created.

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -354,6 +354,160 @@
...
+                "\n        switch(options)",
+                "\n        {",
+                "\n            case FileImportOptions fio:",
+                "\n                var importFile = new ImportFile(Guid.NewGuid());",
+                "\n                try",
+                "\n                {",
+                "\n                    importFile = await (new ImportFile(fio, ImportVariable, Session)).InitializeImportDataAsync() as ImportFile;",
+                "\n                    activity = activity with {SourceId = importFile.Id};",
+                "\n                }",
+                "\n                catch (Exception)",
```
dtrzesniak, line 380: Shouldn't you also record in the activity the message of the Exception?

Something like

```
catch (Exception ex)
{
   importSucceeded = false;
   activity = activity with {SourceId = null, Exception = ex}; // we can see what we want to save, but at least ex.Message.
}
```

dtrzesniak (approved):

ifrs17/Utils/ActivityLog.ipynb
dcolleoni: let's use WithActivityLog
Discussed with Daniel. 
Thanks

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -354,6 +354,160 @@
...
+                "\n                {",
+                "\n                    importSucceeded = false;",
+                "\n                    activity = activity with {SourceId = null};",
+                "\n                }",
+                "\n                activity = activity with {Category = \"Import from File\"};",
+                "\n                if (importSucceeded) await DataSource.UpdateAsync<ImportFile>(importFile.RepeatOnce());",
+                "\n                break;",
```
dtrzesniak, line 387: Can't you simplify this:
1. Add the "await DataSource.UpdateAsync<ImportFile>(importFile.RepeatOnce());"," inside the try
2. Remove importSucceeded,
3. Add "activity = activity with {Category = \"Import from File\"};"," inside the try and catch --> you are creating activity twice (once to set SourceId, once to set Category).
4. Declare importFile inside the try
5. Extract the try/catch as method of type <T> and pass the parameters to avoid the repetition of this code.

I have not thought it through, but I guess this should work and make the code leaner

dcolleoni: the pattern matching can then be used instead of the switch case.

dcolleoni: I suggest to close this PR as is to give clients the possibility to adapt to the new api. 
Improvements in the code can be done in a separate pr. 
The comments are captured in #167

dcolleoni (approved):

PR #181, opened 2023-01-19
Export scenario column

tkukuljan:
The export had duplicate Scenario column, which shouldn't be so.


tkukuljan:
The way I understand the current DataModel is that the Scenario is a property of the ReportingNode. If this is so, I would keep the column Scenario not in Main, but in one of the tables of datanodes of one of the mandatory files for datanodes. 
How should it be done?
@dcolleoni

dcolleoni (approved):

PR #189, opened 2023-01-23
Improve logs v3

andrey-katz-systemorph:
Moves the entire try - catch into a separate generic method. Writes down the switch statement in the ExecuteAsync in a concise manner and adds necessary constructors that are used in ReportInputAndUpdateActivityAsync method. To the import classes added methods WithSession, WithOptions, WithSetReader, to use in case these values are not set by a contructor.


amuolo (changes requested):
the extension methods WithActivityLog require every time the ISessionVariable, IDataSource, and IDataSetImportVariable to be provided. This makes the API lengthy and repetitive. 
I recommend to: 
 - take dataSource from the builder, 
 - move the session and the importVariable to static properties of the newly defined ImportBuilderWriter and ExportBuilderWriter classes. At the end of this notebook, these properties will be assigned to Session and DataSetReader, respectively.

amuolo:
Why are File and Activity two separate classes/tables? Is there a reason to prefer this design better than the simpler approach of having only one class with both info? In the presentation of this feature you anyway merged the information. 
(I am not asking about feasibility, technicalities of which implementation might be coded. I am referring to the chosen code design.)

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -424,82 +483,46 @@
...
+                "\n            DataSetImportOptions dsio => await ReportInputAndUpdateActivityAsync<DataSetImportOptions, ImportDataSet>(log, dsio, \"Import fro Data Set\"),",
+                "\n            _ => null,",
+                "\n        };",
+                "\n        if (activity is null) throw new Exception(\"Import Options object is not an instance of an appropriate class.\");",
                 "\n        await DataSource.UpdateAsync<ImportExportActivity>(activity.RepeatOnce());",
                 "\n        await DataSource.CommitAsync();",
                 "\n        return log;",
                 "\n    }",
+                "\n",
+                "\n    private async Task<ImportExportActivity> ReportInputAndUpdateActivityAsync<TOptions, TImport>(ActivityLog log, ",
```
amuolo, line 500: pls fix the indentation of this method

andrey-katz-systemorph:
> Why are File and Activity two separate classes/tables? Is there a reason to prefer this design better than the simpler approach of having only one class with both info? In the presentation of this feature you anyway merged the information. (I am not asking about feasibility, technicalities of which implementation might be coded. I am referring to the chosen code design.)

I would say, there is a reason. Those are two conceptually different object, and I would not merge them in the DB. First, activity might or might not come with a file. In fact, there can be 4 different objects that are imported: Files, Strings, Streams and DataSets. As you see, all of them populate 4 different tables, they have different properties that characterize them, therefore damping them all into one table would just be a mess, and this actually defines the hierarchy of the abstraction layers in the file. In the presentation I showed the merger of these tables because this is what the client wanted in this particular case, but this infrastructure is supposed to tackle a much wider range of activities, therefore the split into different tables.

andrey-katz-systemorph:
> the extension methods WithActivityLog require every time the ISessionVariable, IDataSource, and IDataSetImportVariable to be provided. This makes the API lengthy and repetitive. I recommend to:
> 
> * take dataSource from the builder,

@amuolo , this is not possible, because this property of the builder is private protected, and no suitable public method exists to fetch it out. This issue was discussed with the developers team, there are reasons  why it is not public and they are very reluctant of opening it for public use.

amuolo:
> > Why are File and Activity two separate classes/tables? Is there a reason to prefer this design better than the simpler approach of having only one class with both info? In the presentation of this feature you anyway merged the information. (I am not asking about feasibility, technicalities of which implementation might be coded. I am referring to the chosen code design.)
> 
> @amuolo I would say, there is a reason. Those are two conceptually different object, and I would not merge them in the DB. First, activity might or might not come with a file. In fact, there can be 4 different objects that are imported: Files, Strings, Streams and DataSets. As you see, all of them populate 4 different tables, they have different properties that characterize them, therefore damping them all into one table would just be a mess, and this actually defines the hierarchy of the abstraction layers in the file. In the presentation I showed the merger of these tables because this is what the client wanted in this particular case, but this infrastructure is supposed to tackle a much wider range of activities, therefore the split into different tables.

Ok I see. Then I agree with you, better to keep the classes separate.

amuolo:
> > the extension methods WithActivityLog require every time the ISessionVariable, IDataSource, and IDataSetImportVariable to be provided. This makes the API lengthy and repetitive. I recommend to:
> > 
> > * take dataSource from the builder,
> 
> @amuolo , this is not possible, because this property of the builder is private protected, and no suitable public method exists to fetch it out. This issue was discussed with the developers team, there are reasons why it is not public and they are very reluctant of opening it for public use.

What we can do at this point is to use static properties and assign to them the global variables Session, DataSource and DataSetReader. In this way we can improve the API.

andrey-katz-systemorph:
> the extension methods WithActivityLog require every time the ISessionVariable, IDataSource, and IDataSetImportVariable to be provided. This makes the API lengthy and repetitive. I recommend to:
> 
> * take dataSource from the builder,
> * move the session and the importVariable to static properties of the newly defined ImportBuilderWriter and ExportBuilderWriter classes. At the end of this notebook, these properties will be assigned to Session and DataSetReader, respectively.

I agree that this is a better solution, and in fact much better in the long term than we have right now. This would lead to a minor change in API, something that was discouraged at this point, but I think that we can still do it. @dcolleoni , do you agree we can still afford ourselves a minor API change here?

dcolleoni:
@andrey-katz-systemorph  pls create a new issue with the improvement previously suggested by Daniel and those arising from your discussion with @amuolo. Assign release 1.1.x to the new issue. 
The issue #167 is meant for release 1.1.0 and it is completed with the scope required by that release. 
Thanks.

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -157,6 +157,11 @@
                 "\n    public string Format {get; init;}",
                 "\n    protected IDataSetImportVariable DataSetReader {get; set;}",
                 "\n    protected ISessionVariable Session {get; set;}",
+                "\n",
+                "\n    public KeyedImportExport() ",
```
dcolleoni, line 161: what is this?

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -329,6 +363,8 @@
                 "\n{",
                 "\n    public string Content {get; init;}",
                 "\n",
+                "\n    public ImportString() : base(){;}",
```
dcolleoni, line 366: maybe you can explain this to me in a call

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -415,92 +474,59 @@
...
-                "\n                break;",
-                "\n",
-                "\n            default:",
-                "\n                throw new Exception(\"Import Options object is not an instance of an appropriate class.\");",
-                "\n                break;",
-                "\n        }",
+                "\n            FileImportOptions fio => await ReportInputAndUpdateActivityAsync<FileImportOptions, ImportFile>(log, fio, \"Import from File \"),",
+                "\n            StringImportOptions sgio => await ReportInputAndUpdateActivityAsync<StringImportOptions, ImportString>(log, sgio, \"Import from String\"),",
+                "\n            StreamImportOptions smio => await ReportInputAndUpdateActivityAsync<StreamImportOptions, ImportStream>(log, smio, \"Import from Stream\"),",
+                "\n            DataSetImportOptions dsio => await ReportInputAndUpdateActivityAsync<DataSetImportOptions, ImportDataSet>(log, dsio, \"Import fro Data Set\"),",
```
dcolleoni, line 492: fro => from.

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -415,92 +474,59 @@
...
+                "\n                                .WithOptions(options)",
+                "\n                                .InitializeImportDataAsync() as TImport;",
+                "\n            activity = activity with {SourceId = import.Id, ",
+                "\n                                    Category = categoryMessage};",
+                "\n            await DataSource.UpdateAsync<TImport>(import.RepeatOnce());",
+                "\n        }",
+                "\n        catch (Exception e)",
+                "\n        {",
+                "\n            activity = activity with {SourceId = null, ",
+                "\n\t\t\t\t\t\t\t\t\tCategory = categoryMessage,",
```
dcolleoni, line 520: something look strange here with indentation?

ifrs17/Utils/ActivityLog.ipynb
andrey-katz-systemorph: This is actually a default constructor that will be used without change by the inheriting classes.

dcolleoni (approved):

amuolo (approved):

PR #215, opened 2023-02-09
Import With Define Format Fix For With Target Option

nnikolopoulos:
**Cashflow** and **Actual** imports uses DataSource as a default target, whereas the target itself should be adjustable via the `.WithTarget( ... )` option.


ifrs17/Import/Importers.ipynb
```diff
@@ -1404,7 +1404,7 @@
                 "\n        log = log.Merge(await ComputeAsync(args, workspace, workspaceToCompute, false)); ",
                 "\n        if(log.Errors.Any()) return Activity.Finish().Merge(log);",
                 "\n    }",
-                "\n    await workspaceToCompute.CommitToTargetAsync(DataSource);",
+                "\n    await workspaceToCompute.CommitToTargetAsync(options.TargetDataSource);",
```
dcolleoni, line 1407: there are other places where DataSource is used. Should they use DataSource or the TargetDataSource?

nnikolopoulos: I only found it relevant to substitute `DataSource` with `options.TargetDataSource` in `workspaceToCompute.CommitToTargetAsync(...)`.

I left the `workspaceToCompute.Initialize( x => x.FromSource(DataSource)` untouched since `ws1` is initialized from `DataSource` in SequenceImportTest (both for Actuals and Cashflow importers).

nnikolopoulos: Having a second look `DataSource` was used by default in two places in `Import.DefineFormat( ... )` for **Cashflows** and **Actuals**:

- `workspaceToCompute.Initialize(x => x.FromSource(DataSource));`
- `await workspaceToCompute.CommitToTargetAsync(DataSource);`

It makes sense to go ahead with this proposed suggestion to use:
`await workspaceToCompute.CommitToTargetAsync(options.TargetDataSource);`
because whenever the importer for **Cashflows** and **Actuals** is called the `Workspace` is initialized from the `DataSource` prior to the call of the importers.

nnikolopoulos: Course of action was discussed internally.

dcolleoni (approved):

PR #229, opened 2023-03-07
Coverage parameter scenario

amuolo:
This PR is to introduce the possibility to import reinsurance coverage parameters for different scenarios.


ifrs17/Utils/EqualityComparers.ipynb
```diff
@@ -140,6 +140,35 @@
...
+                "\n",
+                "\n    public bool Equals(DataNodeParameter x, DataNodeParameter y)",
+                "\n    {",
+                "\n        if (x == null && y == null) return true; ",
+                "\n        if (x == null || y == null) return false; ",
+                "\n        if (!(x.Year == y.Year && x.Month == y.Month && x.DataNode == y.DataNode && x.Scenario == y.Scenario)) return false; ",
+                "\n        if (x is SingleDataNodeParameter && y is SingleDataNodeParameter &&",
+                "\n            ((SingleDataNodeParameter)x).PremiumAllocation == ((SingleDataNodeParameter)y).PremiumAllocation) return true; ",
+                "\n        if (x is InterDataNodeParameter && y is InterDataNodeParameter &&",
+                "\n            ((InterDataNodeParameter)x).LinkedDataNode == ((InterDataNodeParameter)y).LinkedDataNode && ",
```
dcolleoni, line 158: DataNode and LinkedDataNode might be inverted but the comparer should return true

ifrs17/Import/Importers.ipynb
```diff
@@ -503,18 +503,33 @@
...
                 "\n    switch(format)",
                 "\n    {",
-                "\n        case ImportFormats.YieldCurve : ",
+                "\n        case ImportFormats.Cashflow:",
+                "\n        case ImportFormats.Actual: ",
+                "\n        case ImportFormats.Opening:",
+                "\n        case ImportFormats.SimpleValue:",
+                "\n        {",
+                "\n            if(args.Scenario != null) return args.RepeatOnce().ToArray();",
+                "\n            var secondaryArgs = await dataSource.Query<PartitionByReportingNodeAndPeriod>() ",
```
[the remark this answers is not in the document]
dtrzesniak: Should it go the parsing storage and cached?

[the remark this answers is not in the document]
amuolo: Yes, I agree. If we used the parsing storage solely to query the database only once (per type), then we would recycle the data without querying multiple times. This is doable but it would require quite a few changes in the importers: both in the parsing storage to ensure that all needed types are queried, and also in these utility helper methods to be able to pass the parsing storage as parameter.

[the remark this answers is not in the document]
amuolo: I suggest to implement these changes in a separate PR. Can I?

ifrs17/Import/Importers.ipynb
```diff
@@ -1180,13 +1174,52 @@
...
+                "\n        if(ApplicationMessage.HasErrors()) return Activity.Finish().Merge(log);",
+                "\n        ",
+                "\n        // Avoid starting the computation if no best estimate cash flow and actuals has ever been imported ",
+                "\n        if(args.Scenario == null) {",
+                "\n            if(!(await options.TargetDataSource.Query<RawVariable>().Where(x => x.Partition == targetPartition).Take(1).ToArrayAsync()).Any() &&",
+                "\n               !(await options.TargetDataSource.Query<IfrsVariable>().Where(x => x.Partition == targetPartition).Take(1).ToArrayAsync()).Any()) continue;",
+                "\n        }",
+                "\n",
+                "\n        // Remove data nodes which are unaffected by the updated yield curves",
+                "\n        var nominals = await options.TargetDataSource.Query<RawVariable>().Where(x => targetDataNodes.Contains(x.DataNode) && (x.Partition == targetPartition || x.Partition == defaultPartition)).ToArrayAsync();",
```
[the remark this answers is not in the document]
amuolo: That's right! Done!

ifrs17/Import/Importers.ipynb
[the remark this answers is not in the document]
amuolo: I created this ticket to this aim
https://github.com/Systemorph/IFRS17CalculationEngine/issues/237

dtrzesniak (approved):

ifrs17/Import/Importers.ipynb
```diff
@@ -503,18 +503,33 @@
...
-                "\n        case ImportFormats.YieldCurve : ",
+                "\n        case ImportFormats.Cashflow:",
+                "\n        case ImportFormats.Actual: ",
+                "\n        case ImportFormats.Opening:",
+                "\n        case ImportFormats.SimpleValue:",
+                "\n        {",
+                "\n            if(args.Scenario != null) return args.RepeatOnce().ToArray();",
+                "\n            var secondaryArgs = await dataSource.Query<PartitionByReportingNodeAndPeriod>() ",
+                "\n                .Where(x => x.ReportingNode == args.ReportingNode && x.Year == args.Year && x.Month == args.Month && x.Scenario != null)",
+                "\n                .Select(x => new ImportArgs(x.ReportingNode, x.Year, x.Month, default(Periodicity), x.Scenario, ImportFormats.Cashflow)).ToArrayAsync();",
```
dcolleoni, line 519: ImportFormat is set to ImportFormats.Cashflow in every case? also when ImportFormats is Actual and Opening?

ifrs17/Import/Importers.ipynb
```diff
@@ -503,18 +503,33 @@
...
+                "public async Task<ImportArgs[]> GetAllArgsAsync(ImportArgs args, IDataSource dataSource, string format)",
                 "\n{",
                 "\n    ImportArgs[] allArgs;",
                 "\n    switch(format)",
                 "\n    {",
-                "\n        case ImportFormats.YieldCurve : ",
+                "\n        case ImportFormats.Cashflow:",
+                "\n        case ImportFormats.Actual: ",
+                "\n        case ImportFormats.Opening:",
+                "\n        case ImportFormats.SimpleValue:",
```
dcolleoni, line 514: why SimpleValue is here? this format does not compute anything.

dtrzesniak: Good point.

amuolo: yes, agree, we can remove simple value.

ifrs17/Import/Importers.ipynb
amuolo: I believe this is correct because the secondary args are all those scenario args that come into play when importing something Best Estimate. In this implementation, scenario calculations have import format cashflow.

amuolo: This line has not been changed by this PR, that is, same behavior was there prior to this PR.

ifrs17/Import/Importers.ipynb
```diff
@@ -1180,13 +1173,52 @@
...
+                "\n",
+                "\n    if(log.Errors.Any()) return Activity.Finish().Merge(log); ",
+                "\n    var toCommitParameters  = (await workspace.Query<DataNodeParameter>().ToArrayAsync()).Except(committedParameters, ParametersComparer.Instance());",
+                "\n    if (!toCommitParameters.Any()) {",
+                "\n        ApplicationMessage.Log(Warning.VariablesAlreadyImported);   ",
+                "\n        return Activity.Finish().Merge(log);",
+                "\n    }",
+                "\n",
+                "\n    var allArgs = await GetAllArgsAsync(primaryArgs, options.TargetDataSource, ImportFormats.DataNodeParameter);",
+                "\n    var targetDataNodes = toCommitParameters.Select(x => x.DataNode).ToHashSet();",
```
dcolleoni, line 1194: For InterDataNodeParameter you also have a linkedDataNode. Why is this not a targetDataNode?

ifrs17/Import/Importers.ipynb
dcolleoni: Having FormatCashflow computes from RawVar and it does not compute Actuals such as Advance and Overdue. 
If you re-import an Actual BE and you have already imported an Actual scenario you need to recompute the scenario performing calculation of Advance and Overdue. Now I think you would re-compute them.

amuolo: Do you suggest to compute from Actual in case of a (re-)import of an Actual BE?

ifrs17/Import/Importers.ipynb
amuolo: Because the linked data nodes are retrieved from the secondary scope of the import storage employing the relaxed logic. I think this might be correct, right?

ifrs17/Import/Importers.ipynb
amuolo: ok, I am convinced and I agree with you: the import format in the case of BE actuals need to be ImportFormats.Actual. I have applied the change to achieve this behavior in the lines you picked up.

ifrs17/Import/Importers.ipynb
amuolo: You are super right! If I do not consider also the LinkedDataNode and upload them too to the workspace, the import storage wouldn't have a chance to retrieve the corresponding IfrsVariables or RawVariables. Now I fixed this. Thanks!

ifrs17/Import/Importers.ipynb
```diff
@@ -503,18 +503,33 @@
...
                 "\n    ImportArgs[] allArgs;",
                 "\n    switch(format)",
                 "\n    {",
-                "\n        case ImportFormats.YieldCurve : ",
+                "\n        case ImportFormats.Cashflow:",
+                "\n        case ImportFormats.Actual: ",
+                "\n        case ImportFormats.Opening:",
+                "\n        {",
+                "\n            if(args.Scenario != null) return args.RepeatOnce().ToArray();",
+                "\n            var importFormat = (format == ImportFormats.Actual)? ImportFormats.Actual : ImportFormats.Cashflow;",
```
dcolleoni, line 516: Now I think Opening also imports Advance and Overdue, and you would then need to leave format = opening else with Cashflow you would have the same problem we discussed for actuals.

dcolleoni:
Why do we have so many changes in BMs? is it because of the format we fixed?

ifrs17/Import/Importers.ipynb
amuolo: ok done

amuolo:
> Why do we have so many changes in BMs? is it because of the format we fixed?

Not only, it is also because of line 570 of the Importers notebook, that is, variables scenarios that are equal to BE are not saved anymore. Before we didn't have this additional filter and we duplicated the values.

ifrs17/Import/Importers.ipynb
```diff
@@ -503,18 +503,32 @@
...
+                "\n                         .Distinct();",
                 "\n                var targetPartitions = await dataSource.Query<PartitionByReportingNodeAndPeriod>()",
-                "\n                    .Where(x => x.Year == args.Year && x.Month == args.Month && !scenariosWithYieldCurves.Contains(x.Scenario)).OrderBy(x => x.Scenario).ToArrayAsync();",
+                "\n                    .Where(x => x.Year == args.Year && x.Month == args.Month && !scenarios.Contains(x.Scenario)).OrderBy(x => x.Scenario).ToArrayAsync();",
                 "\n                var targetScenarios = targetPartitions.Where(x => x.Scenario != null).Select(x => x.Scenario);",
                 "\n                if(targetScenarios.Any()) ApplicationMessage.Log(Warning.ScenarioReCalculations, String.Join(\", \", targetScenarios));",
                 "\n                allArgs = targetPartitions.Select(x => new ImportArgs(x.ReportingNode, x.Year, x.Month, default(Periodicity), x.Scenario, ImportFormats.Cashflow)).ToArray();",
```
dcolleoni, line 534: the format is set to Cashflow here. While for YC I see as necessary, for a generic ParameterUpdated it might not be necessary. Given the parameters we have now it would be enough format Actual. This would speed up import. On the other side the code would get more complex having to decide which format is the best given the scenario. Also, we might add new parameters in the future making the logic even more complex. 
In summary, I think it is a good decision to keep Cashflow format here.

ifrs17/Import/Importers.ipynb
```diff
@@ -1180,13 +1174,53 @@
...
+                "\n        if(ApplicationMessage.HasErrors()) return Activity.Finish().Merge(log);",
+                "\n        ",
+                "\n        // Avoid starting the computation if no best estimate cash flow and actuals has ever been imported ",
+                "\n        if(args.Scenario == null) {",
+                "\n            if(!(await options.TargetDataSource.Query<RawVariable>().Where(x => x.Partition == targetPartition).Take(1).ToArrayAsync()).Any() &&",
+                "\n               !(await options.TargetDataSource.Query<IfrsVariable>().Where(x => x.Partition == targetPartition).Take(1).ToArrayAsync()).Any()) continue;",
+                "\n        }",
+                "\n",
+                "\n        // Only nominals corresponding to the target data nodes are added to the workspace",
+                "\n        var nominals = await options.TargetDataSource.Query<RawVariable>().Where(x => targetDataNodes.Contains(x.DataNode) && (x.Partition == targetPartition || x.Partition == defaultPartition)).ToArrayAsync();",
```
dcolleoni, line 1214: don't you need also previous period partition?
This solution seems to me a bit complex... and it is too easy to miss some pieces... also hard to maintain...

ifrs17/Import/Importers.ipynb
amuolo: Yes, thanks! I also think it is quite a general rule that would remain valid for more parameters.

ifrs17/Import/Importers.ipynb
amuolo: previous partition added! We can improve this logic with the next PR.

dcolleoni (approved):

PR #268, opened 2023-04-11
Convert credit default rate to monthly basis

dcolleoni:
Convert to monthly basis (same as CF) and update bms


ifrs17/Import/ImportStorage.ipynb
```diff
@@ -363,7 +363,7 @@
                 "\n        // if Partner == Internal then return 0;",
                 "\n        if(!PartnerRating.TryGetValue(dataNodeData.Partner, out var rating))                   ApplicationMessage.Log(Error.RatingNotFound, dataNodeData.Partner);",
                 "\n        if(!CreditDefaultRates.TryGetValue(rating[period].CreditRiskRating, out var rate))     ApplicationMessage.Log(Error.CreditDefaultRateNotFound, rating[period].CreditRiskRating);",
-                "\n        return rate[period].Values[0];",
+                "\n        return Math.Pow(1 + rate[period].Values[0], 1d / 12d) - 1d;",
```
dtrzesniak, line 366: add a d also here, 1d +...

dtrzesniak: Good, but I would add some documentation, that you can write based on this: https://calculate.onl/convert-annual-interest-rates/

I think we do not need any other granularity (i.e. quarterly), as we always do all calculations on monthly basis. So we should be good with your formula.

dtrzesniak (approved):

Issue #266, opened 2023-04-10
Refactor activity log db Tables


dcolleoni:
Reduce the number of tables to record activity log in db

PR #273, opened 2023-04-19
Refactor ActivityLog DataBase Tables

nnikolopoulos:
Introduce IOActivity and IOContent tables (this PR closes issue #266).
- Rename ImportExportActivity to IOActivity.
- Rename KeyedImportExport to IOContent and make it non-abstract.

The proposed changes result to a new table (IOContent) in the DB which should contain information that previously existed in the tables: ExportFile, ImportFile, ImportString, ImportDataSet and ImportStream (IO Content replaces the aforementioned tables).


dcolleoni (approved):
Let's have a meeting where you show me the DB with data

ifrs17/Utils/ActivityLog.ipynb
```diff
@@ -150,7 +150,7 @@
         {
             "cell_type": "code",
             "source": [
-                "public abstract record KeyedImportExport : KeyedRecord",
+                "public record IOContent : KeyedRecord",
                 "\n{",
```
andrey-katz-systemorph, line 154: Do we really want to see the table of IO content? We have separate tables of imports and exports that we do not remove. Hence I believe this record must be abstract.

nnikolopoulos: @andrey-katz-systemorph I was under the impression that the a new collective table (IOContent) should replace the following:

ExportFile, ImportFile, ImportDataSet, ImportString, ImportStream

and contain all their columns. 

Making it non-abstract actually drops these 5 tables from the DB and creates a new one with the combined information.

andrey-katz-systemorph: Then I do not understand this PR at all. You are still keeping all these record, so these tables will be formed, like or dislike. Moreover, from reading the code you populate *these tables* rather than your IOContent. Also you have abstract KeyedImport that inherits from non-abstract IOContent, which looks unusual (not sure whether it is technically allowed, but it breaks the logical hierarchy of the records.)

andrey-katz-systemorph (changes requested):
OK, I went through all the files and in my humble opinion it does not do what you would like it to do and the problems are actually multifold. 

1. You generate upon migration tables for any non abstract class, there is no guarantee that they are populated properly though. 
2. At the moment you call .WithActivityLog() on the import, you generate ImportBuilderWriter. Take a look how the ExecuteAsync() works. It calls the ReportInputAndUpdateActivityAsync and writes the result in the TImport table, which is arguably not what you want.
3. If you want the IOContent to be populated, you should get rid all the records that you do not want, the import inside the  ReportInputAndUpdateActivityAsync should be of IOContent type (and TImport becomes obsolete all together) and consequently move all the logic that is now defined at the lower layers up all the way to the IOContent. There might be many variations to this "recipee" but I believe this roughly the way we should move.

ifrs17/Utils/ActivityLog.ipynb
nnikolopoulos: To test if these changes have the expected behavior I used the `Scaffold` notebook and replaced the contents of `Initial` to have the new tables in the DB (+IOContent +IOActivity -ImportExportActivity -ImportFiles -ImportString -ImportStream -ImportDataSet). 

Then I Initialized the DB and IOContent table was populated:
image

I didn't push yet here the updated migration notebooks.

Nevertheless I will try to implement your requested changes because I agree that the logical hierarchy currently is not ideal @andrey-katz-systemorph.

nnikolopoulos:
Following discussion with @dcolleoni I tried to avoid a major restructuring of the ActivityLog notebook.

Therefore I did not proceed with the requested changes by @andrey-katz-systemorph.

The latest commits: 

- move Name and ContentType properties from ImportFile and ExportFile records all the way up to the IOContent record and also
- update the migration notebooks.

@andrey-katz-systemorph the resulting IOContent table in the DataBase taking these changes into account looks like this after some mock imports / exports:

image

image

Currently the abstract modifier in the KeyedImport record is used to prevent the creation of a KeyedImport table in the DB.

@dtrzesniak perhaps you can also review the changes at this point.

andrey-katz-systemorph (approved):

ifrs17/Utils/ActivityLog.ipynb
dtrzesniak: @andrey-katz-systemorph No, the tables will not be created, rather there will be 1 table only in the DB, as this is the concrete record. I think this is what we want.

dtrzesniak: abstract inheritance in the middle should also be ok, as abstract is only thereto avoid code repetition.

dtrzesniak (approved):

PR #296, opened 2023-05-04
Ey cases

### LINKED ISSUE #297: Fix to YC and AM
We fix the behavior of the YC when the values are shorter than expected (was adding 0, should be repeating last value)

We fix the Am factor for LIC and Run-off cases (under certain extreme cases it would go above 1 and make a switch)

Fixed by #296

dtrzesniak:
This PR fixes issues with AM factor and Yield Curves.
It also add many practical cases and examples, which are not documented yet, but have the corresponding data already imported.

Fixes #297


ifrs17-template/Constants/CalculationEngine.ipynb
```diff
@@ -22,7 +22,8 @@
                 "var projectName = \"ifrs17\";",
                 "\nvar environmentName = \"v1.3.0\";",
                 "\nvar notebookName = \"CalculationEngine\";",
-                "\nvar calculationEngine = $\"#!import \\\"//{projectName}/{environmentName}/{notebookName}\\\"\";"
+                "\nvar calculationEngine = $\"#!import \\\"//{projectName}/{environmentName}/{notebookName}\\\"\";",
```
dtrzesniak, line 25: To be reset once

ifrs17-template/Test/PracticalCase-CompareReinsuranceTest.ipynb
```diff
@@ -84,7 +84,7 @@
         {
             "cell_type": "code",
             "source": [
-                "bops.Where(x => x.GroupOfContract==grossBusinessReinsured && x.Scenario==null).Sum(x=>x.Value).Should().BeApproximately(-100.00828, Precision);"
+                "bops.Where(x => x.GroupOfContract==grossBusinessReinsured && x.Scenario==null).Sum(x=>x.Value).Should().BeApproximately(-100.05818, Precision);"
```
dcolleoni, line 87: Why is bop changing?
I expect only the AM to change due to the fix of AMF > 1.

ifrs17/Import/2ImportScope-PresentValue.ipynb
```diff
@@ -890,13 +897,13 @@
...
+                "\n                            .DefaultIfEmpty()",
                 "\n                            .Aggregate(1d, (x, y) => x * y);",
+                "\n                            ",
                 "\n    [NotVisible] string EconomicBasis => GetContext();",
                 "\n",
                 "\n    string EstimateType => EstimateTypes.F;",
-                "\n    double Value => Math.Abs(amortizedFactor - 1d) > Precision",
-                "\n                        ? 1d - amortizedFactor",
-                "\n                        : 1d;",
+                "\n    double Value => 1d - amortizedFactor;",
```
dcolleoni, line 906: The run off case would produce a MonthlyAmortizationFactors of 0 and length 1. Right? 
So that Value = 1.

dcolleoni:
Do you need to increase the memory required to run the Template Test?
If yes, could you update the recommendation in the markdown of that NB? 
Thanks

dcolleoni (approved):

ifrs17/Import/2ImportScope-PresentValue.ipynb
dtrzesniak: .DefaultIfEmpty() will give a double[]{0d} if the skip().take() gives an empty. So, yes, it will producea MonthlyAmortizationFactors of 0 and length 1, thus Value of 1. All correct.

Issue #307, opened 2023-05-17
LINQ Translation Error in the Reports


nnikolopoulos:
**Description**
Reports inside the `Reports` notebook cannot run because of a LINQ translation error. 

To reproduce:
- Run the `Reports` notebook.
- Use the data present in the Database. 

**Screenshots**
<img width="1920" alt="image" src="https://github.com/Systemorph/IFRS17CalculationEngine/assets/114566025/ed3e1e04-a988-4b2f-8920-27e9551c45b4">

**Additional context**
An update of the `Initial` notebook will be necessary, after the latest (small) changes in the DataModel.

PR #308, opened 2023-05-17
LINQ Translation in Reports

nnikolopoulos:
Adjust `ReportConfigurationAndUtlis` and update `Initial` notebooks. 
This PR fixes issue #307.


ifrs17/Report/ReportConfigurationAndUtils.ipynb
dcolleoni, line 1: Is this change to support physical DB?

nnikolopoulos: Yes, this change is meant to fix the Reports in case of Initialization from a physical DB.

Issue #240, opened 2023-03-16
Warning message at input parsing


dcolleoni:
Log a warning when the expected type of a the content of a column does not match the expectation. 
Example of this is AccidentYear in CF,ACT,OP,SimpleImport: 
Expected input of type int, if a double is provided the value is not parsed without any warning to the user.

nnikolopoulos:
A PR for this issue has been here #316 targeting a feature branch (**ExternalStakeholderPostProd4**) which will eventually be merged to main (and develop).

PR #316, opened 2023-05-25
Log an Error for Invalid AY Types

nnikolopoulos:
Log an error when the input of `AccidentYear` does not match the expected type (int).
This PR closes issue #240.


ifrs17/Constants/Validations.ipynb
```diff
@@ -122,6 +123,7 @@
                 "\n    (Error.AocTypeCompulsoryNotFound      , _) => $\"Not all compulsory AoC Types have been imported.\",",
                 "\n    (Error.AocTypePositionNotSupported    , 1) => $\"The position of the AoC Type {s[0]} is not supported.\",",
                 "\n    (Error.AocConfigurationOrderNotUnique , _) => $\"Two or more AoC Configurations have the same Order.\",",
+                "\n    (Error.AccidentYearTypeNotValid       , 1) => $\"The parsed AccidentYear type {s[0]} is invalid. Expected Accident Year input of type int.\",",
```
dcolleoni, line 126: The parsed AccidentYear type {s[0]} is invalid. => 
The parsed AccidentYear {s[0]} is invalid.

ifrs17/Import/Importers.ipynb
```diff
@@ -1369,13 +1369,20 @@
                 "\n                if(values.Length == 0 && !parsingStorage.MandatoryAocSteps.Contains(new AocStep(aocType, novelty))) return null;",
                 "\n            }",
                 "\n            ",
+                "\n            var accidentYearInput = (datarow.Field<string>(nameof(RawVariable.AccidentYear)));",
```
dcolleoni, line 1372: do this only inside the if(hasAccidentYearColumn)

ifrs17/Import/Importers.ipynb
```diff
@@ -1369,13 +1369,20 @@
...
+                "\n            var isAccidentYearInt = Int32.TryParse(accidentYearInput, out var accidentYear);",
+                "\n            if(hasAccidentYearColumn)",
+                "\n            { ",
+                "\n                if(!(isAccidentYearInt || accidentYearInput == null)) ApplicationMessage.Log(Error.AccidentYearTypeNotValid, accidentYearInput);",
```
dcolleoni, line 1376: transform this condition to be a positive condition: 
!isAYint && AYInput != null

ifrs17/Import/Importers.ipynb
nnikolopoulos: Can't move the variables inside the if because the will not be accessible again in line 1385

dcolleoni: Define accidentYear outside with a default value.
Then do the checks and inside the if.

ifrs17/Import/Importers.ipynb
```diff
@@ -1369,13 +1369,20 @@
                 "\n                if(values.Length == 0 && !parsingStorage.MandatoryAocSteps.Contains(new AocStep(aocType, novelty))) return null;",
                 "\n            }",
                 "\n            ",
+                "\n            var accidentYearInput = (datarow.Field<string>(nameof(RawVariable.AccidentYear)));",
+                "\n            var isAccidentYearInt = Int32.TryParse(accidentYearInput, out var accidentYear);",
+                "\n            if(hasAccidentYearColumn)",
+                "\n            { ",
+                "\n                if(!isAccidentYearInt && accidentYearInput != null) ApplicationMessage.Log(Error.AccidentYearTypeNotValid, accidentYearInput);",
```
dcolleoni, line 1376: return null in this case. 
You don't need an explicit variable for "isAccidentYearInt"

ifrs17/Import/Importers.ipynb
```diff
@@ -1369,13 +1369,24 @@
                 "\n                if(values.Length == 0 && !parsingStorage.MandatoryAocSteps.Contains(new AocStep(aocType, novelty))) return null;",
                 "\n            }",
                 "\n            ",
+                "\n            var accidentYearInput = default(string);",
```
dcolleoni, line 1372: let's have a call.

dcolleoni (approved):

PR #327, opened 2023-06-23
Contract Term

sbatka:
- add Contract Term to DataModel, defined at GroupOfContract level in Months
- add Contract Term to DataNodeData constructor
- ContractTerm can be used to calculate deferrable or amortized positions of Balance Sheet


ifrs17/Import/ImportStorage.ipynb
```diff
@@ -58,7 +58,7 @@
             "cell_type": "code",
             "source": [
                 "#!import \"../Utils/EqualityComparers\"",
-                "\n#!import \"../Utils/ImportCalculationMethods\"",
+                "\n//#!import \"../Utils/ImportCalculationMethods\"",
```
dcolleoni, line 61: why is this commented out?

dcolleoni (approved):

PR #356, opened 2023-07-27
Little things on projections

dtrzesniak:
Added a fix for bundled projections
Added a report example
Some other minor fixes


dcolleoni:
What about replacing the old input file with the new (long projection). I think it is important to have a test on these projected values.

dcolleoni (approved):

PR #362, opened 2023-08-02
359 extendable warnings and errors

andrey-katz-systemorph:
Extends the enumerables of the errors and warnings to the classes in a minimal possible way in order not to break the existing code, allowing however the inheritance by external projects. The only caveat of this approach is the way switch is going to work (when must be invoked).


ifrs17/Constants/Validations.ipynb
```diff
@@ -45,14 +45,9 @@
...
-            "cell_type": "code",
+            "cell_type": "markdown",
             "source": [
-                "public enum Warning {",
-                "\n    // Import",
-                "\n    ActiveDataNodeWithCashflowBOPI, VariablesAlreadyImported, VariablesAlreadyCalculated, ScenarioReCalculations, MandatoryAocStepMissing,",
-                "\n    // Default",
-                "\n    Generic",
-                "\n}; "
+                "Comment: Warnings and Errors are designed to act similar enumerables, that allow however inheritance in the client projects. However, this stricture is not identical to the enumerables. An important difference is that it does not allow using them directly in the *case* statements. The workaround would be to use the InternalValue, which is set to be the string name of the variable itself. Also, we do not keep the numerical value, attached to the enumerable, because we do not use it anywhere. "
```
tkukuljan, line 50: This paragraph could be written better

dcolleoni: Let's remove the comment altogether.

ifrs17/Constants/Validations.ipynb
```diff
@@ -61,45 +56,15 @@
...
-            "metadata": {},
-            "execution_count": 0,
-            "outputs": []
-        },
-        {
-            "cell_type": "markdown",
-            "source": [
-                "# Messages"
+                "public abstract class ValidationBase{",
+                "\n    public string InternalValue {get; protected set;}",
```
dcolleoni, line 60: Instead of InternalValue I prefer ErrorCode : rename every occurrence, thanks!

ifrs17/Constants/Validations.ipynb
```diff
@@ -108,92 +73,30 @@
...
-                "\n    (Error.InvalidDataNode                , 1) => $\"Data imported for invalid Data Node {s[0]}.\",",
-                "\n    (Error.InvalidDataNodeForOpening      , 1) => $\"Data imported for invalid Data Node or for a Data Node after its inception year {s[0]}.\",",
-                "\n    (Error.InvalidCashFlowPeriodicity     , 1) => $\"Invalid CashFlowPeriodicity parameter for Data Node {s[0]}.\",",
-                "\n    (Error.MissingInterpolationMethod     , 1) => $\"Missing InterpolationMethod parameter for Data Node {s[0]}.\",",
-                "\n    (Error.InvalidInterpolationMethod     , 1) => $\"Invalid InterpolationMethod parameter for Data Node {s[0]}.\",",
-                "\n    (Error.InvalidEconomicBasisDriver     , 1) => $\"Invalid EconomicBasisDriver parameter for Data Node {s[0]}.\",",
-                "\n    (Error.InvalidReleasePattern          , 1) => $\"Invalid ReleasePattern parameters for Data Node {s[0]}.\",",
+                "public class Warning : ValidationBase{",
+                "\n",
+                "\n    protected const string DefaultMessage = \"Warning not found.\";",
```
dcolleoni, line 78: I would remove this variable and move the message in line 96

ifrs17/Constants/Validations.ipynb
```diff
@@ -202,16 +105,193 @@
...
-                "\n    (Warning.VariablesAlreadyImported       , 0) => $\"The import of the current file does not contain any new data. Hence, no data will be saved or calculations will be performed.\",",
-                "\n    (Warning.MandatoryAocStepMissing        , 3) => $\"The AoC step ({s[0]}, {s[1]}) is not imported for ({s[2]}).\",",
-                "\n    (Warning.ScenarioReCalculations         , 1) => $\"The present Best Estimate import makes the result of dependent Scenarios out of date. Hence, the following Scenarios are re-calculated: {s[0]}.\", ",
-                "\n    // Default",
-                "\n    (Warning.Generic                        , _) => $\"{s[0]}\",",
-                "\n    (_                                      , _) => $\"Warning not found.\"",
-                "\n};"
+                "public class Error : ValidationBase{",
+                "\n",
+                "\n    protected const string DefaultMessage = \"Error not found.\";",
```
dcolleoni, line 110: Same here: remove variable and move message in the switch.

dcolleoni (approved):

ifrs17/Constants/Validations.ipynb
andrey-katz-systemorph: Removed.

ifrs17/Constants/Validations.ipynb
andrey-katz-systemorph: It is a little bit confusing name, because this variable applies both to the error and the warning. How about MessageCode or InternalCode or simply Code?

ifrs17/Constants/Validations.ipynb
andrey-katz-systemorph: I would prefer to leave it because I use this variable in inheriting classes (client projects). This is the way to avoid bugs that would stem from the typos. I can define a generic constant with this name, but there should be somewhere an accessible variable with this string value.

ifrs17/Constants/Validations.ipynb
andrey-katz-systemorph: Same as in the comment before: it is used more than once.

dcolleoni:
Dear @andrey-katz-systemorph , 
pls add a correction to the CalculationEngine NB in template/Constant => variable calculationEngine is currently wrong in develop branch.

Thanks.

ifrs17/Constants/Validations.ipynb
dcolleoni: I like MessageCode. 
Thanks, I didn't realize the confusion...

ifrs17/Constants/Validations.ipynb
dcolleoni: ok, got it. Thanks. 
I agree let's keep it!

ifrs17/Constants/Validations.ipynb
andrey-katz-systemorph: Changed to MessageCode everywhere

andrey-katz-systemorph:
> Dear @andrey-katz-systemorph , pls add a correction to the CalculationEngine NB in template/Constant => variable calculationEngine is currently wrong in develop branch.
> 
> Thanks.

I reverted it to what we currently have on the prod.

PR #368, opened 2023-08-10
Include the current yield curves  

andrey-katz-systemorph:
Current CE does queries the current curves if the locked-in is not present (as might be the case with the VFA valuation). This PR solves this bug.


dcolleoni (approved):

dcolleoni:
@andrey-katz-systemorph, pls create a bug in the board and link it to this PR. Thanks!

PR #406, opened 2023-10-19
Compute function fix for import of Gross without Reins

dcolleoni:
Importing a GIC without the associated GRIC leads to: 
- GRIC in primary Scope => recalculated
- RawVariable are saved with snapshot mode filtering for the primary Scope. 
This leads to the wrong deletion of GRIC RawVariable. 

The filter now excludes those DataNode that are added to the PrimaryScope. 

In the ImportStorage fixed an issue with RawVariable queried from DataSource:
- all dnAddedToPrimaryScope are considered not only those linked to a dn in primary scope. 
This makes the re-calculation of LRC-PAA dn correct even in case of reimport without those input. 

Improved and fixed the ScenarioDataImportTest.


ifrs17/Constants/Enums.ipynb
```diff
@@ -299,7 +299,13 @@
         {
             "cell_type": "markdown",
             "source": [
-                "## Import Scope"
+                "## Import Scope",
+                "\nImport Scope is used in the [ImportStorage](../Import/ImportStorage) and in the [Compute Function](../Import/Importers#calculate-ifrs-variables-for-all-scenarios) to drive the calculation of new IfrsVariables. Each import scope contains a collection of Data Node.",
+                "\n",
+                "\nIts values indicate: ",
+                "\n- <code>Primary</code> : Data nodes for which [IfrsVariable](../DataModel/DataStructure#ifrs-variable) will be calculated.",
+                "\n- <code>AddedToPrimary</code> : Subset of Primary. It contains data nodes linked to data nodes provided in input that need recalculation. ",
```
calderinisystemorph: Since you add it, I think "Subset" is misleading (adding a subset to its own set does not change the set itself). This is more a enrichment. I would describe it as a complementary set or, set of dependent Datanodes

ifrs17/Import/ImportStorage.ipynb
```diff
@@ -279,7 +280,8 @@
                 "\n        ",
                 "\n        //Variables",
                 "\n        var rawVariables = parsedRawVariables.Concat(openingRawVariables)",
-                "\n            .Concat(await querySource.Query<RawVariable>().Where(rv => rv.Partition == TargetPartition).Where(rv => primaryScopeFromLinkedReinsurance.Contains(rv.DataNode)).ToArrayAsync()); ",
+                "\n            .Concat(await querySource.Query<RawVariable>().Where(rv => rv.Partition == TargetPartition)",
+                "\n            .Where(rv => primaryScopeFromLinkedReinsurance.Contains(rv.DataNode)).ToArrayAsync()); ",
```
calderinisystemorph: Why do we add RawVariables from primaryScopeFromLinkedReinsurance but not from primaryScopeForPaaLrc even if they are in AddedPrimaryScope

dcolleoni: This is not part of the changes of this PR. 
The reason for it is that a PAA-LRC GIC might require calculation even though no new inputs are available. 
However, I think it is not correct as it could delete variable in some cases. It is safer to add those RawVariables anyway. 
Thanks for the observation.

ifrs17/Constants/Enums.ipynb
```diff
@@ -299,7 +299,13 @@
...
+                "\n- <code>Secondary</code> : Data nodes that provide inputs to the Data nodes in primary scope. No variable will not be calculated."
```
nnikolopoulos: @dcolleoni in the markdown for secondary scope, do you instead mean "No new variable will be calculated"?

calderinisystemorph (approved):

PR #410, opened 2023-10-26
Ey cases update

dtrzesniak:
Update of EY cases


ifrs17-template/Test/EY-Ifrs17UseCases.ipynb
```diff
@@ -362,9 +362,9 @@
         {
             "cell_type": "code",
             "source": [
-                "bop.Value.Should().BeApproximately(-5d, Precision);",
-                "\nev.Value.Should().BeApproximately(7.5d, Precision);",
-                "\neop.Value.Should().BeApproximately(2.5d, Precision);"
+                "// bop.Value.Should().BeApproximately(-5d, Precision);",
```
dcolleoni, line 365: What is missing to enable these checks? 
I can create an issue to bring these assertions back.

dcolleoni (approved):

ifrs17-template/Test/EY-Ifrs17UseCases.ipynb
dtrzesniak: The feature is still missing, as we discussed, we should not recognize this position under CSM in this specific case... We need to see how to implement these requirements, as I was thinking and was still not fully convinced what we discussed is fully correct.

PR #421, opened 2023-11-13
Opening outside the inception year

### LINKED ISSUE #403: Opening can be imported outside of the inception year
Import of Opening outside the inception year should trigger an error validation.

andrey-katz-systemorph:
Logs an error is an opening is outside of the inception year


ifrs17/Import/Importers.ipynb
```diff
@@ -496,9 +496,15 @@
                 "\n    ",
                 "\n    var dataNodesImported = table.Rows.Select(x => x.Field<string>(nameof(RawVariable.DataNode))).ToHashSet();",
                 "\n    var dataNodesDefined = await targetDataSource.Query<GroupOfContract>().Where(x => dataNodesImported.Contains(x.SystemName)).ToArrayAsync();",
-                "\n    var dataNodeStatesDefined = await targetDataSource.Query<DataNodeState>().Select(x => x.DataNode).ToArrayAsync();",
+                "\n    var dataNodeStatesDefined = await targetDataSource.Query<DataNodeState>().Select(x => new {DataNode = x.DataNode, InceptionYear = x.Year}).ToArrayAsync();",
```
dcolleoni, line 499: Can you add (in the PR or in the issue) a description of what is wrong in the code, and how this implementation fixes it?
Currently, I fail to see it... thanks

andrey-katz-systemorph: Well, let me maybe explain right here. If you carefully examine the code, you will discover that until now we never did this check in the first place, so not surprisingly this import succeeded.

So to fix I chose this place, where we query the data note state anyway. Data node state carries the information about the the inception year, which is a year, so I just expand here the query, to keep the information about the data node and the inception year. Note that the year of the args, that arrive here, carry the information about the opening year, as it comes to us from the opening file. So at the next step, if the import format is opening, I generate an enumerable with all the data nodes that have the opening outside the inception year, and if this enumerable is not empty, log an error for each entry. The data DataNodeFactoryAsync is called by the opening importer at the very early stage (line 9), at line 10 it discovers that the that the activity has errors and exists. Does it make sense?

andrey-katz-systemorph: Move this validation into the parser and exploit the storage data.

dcolleoni:
Closing this PR.

PR #447, opened 2024-03-11
Parameter Import config

dcolleoni:


amuolo:
@dcolleoni how does the discussed test look like? is the ref data import failing with record types with non-trivial key?

fyi: @calderinisystemorph

dcolleoni:
> @dcolleoni how does the discussed test look like? is the ref data import failing with record types with non-trivial key?
> 
> fyi: @calderinisystemorph

The test on ReferenceData using a type with Key passes without specifying the key in the import plugin. 
Tests are now green with delay (which makes sure the sync has the time to occur after the import is completed.)

pekishev (approved):

